PR #230, opened 2015-01-02
Tidy layout & add mustard

matthew-andrews:


views/layout.html
```diff
@@ -64,18 +49,18 @@
...
-		{{#if @root.flags.javascript.isSwitchedOn}}
-		<script src="/{{__name}}/main.js"></script>
 		<script>
-		(function () {
-			var html = document.querySelector('html');
-			html.classList.add('js-success');
-		}())
+		(function (src) {
+			if (cutsTheMustard) {
+				document.documentElement.classList.add('js-success');
```
aintgoin2goa: Why is this being added to html and body?

matthew-andrews: because i'm lazy and copied and pasted it from engels.

consider it gone.

matthew-andrews: oh wait it was there before

PR #232, opened 2015-01-05
Fix quote mark issue in following show more button

adgad:
@matthew-andrews 
Fixes https://app.getsentry.com/nextftcom/ft-next-frontend/group/49087445/


matthew-andrews:
:+1:

adgad:
Updated to fix build problem (search suggestions still depended on reqwest)

PR #234, opened 2015-01-05
Add polyfills

matthew-andrews:
@commuterjoy had some more feedback from Mark about javascript not working on his iOS 7 device…  we need polyfills to support modern browsers…  would you mind temporarily using polyfill.io with a view to replacing this with your proper polyfill solution once it's ready…


commuterjoy:
replaced by the work paul is doing

PR #235, opened 2015-01-06
Fix stream links on following page going to wrong page

adgad:
/cc @matthew-andrews Fixes https://github.com/Financial-Times/next-bugs/issues/29


adgad:
:+1:

PR #252, opened 2015-01-16
Topic cards

adgad:
/cc @wheresrhys 

Probably need a better place for all these svgs, and there's some refactoring that could be done RE streams at some point.
Also currently appearing in a random order because of the sorting (but it's easier to solve that once refactoring a bit later)
topics


wheresrhys:
:+1: We need to look at relationships between next-card/article-card/topic/card/other-cards too. Monster refactor ahead!

PR #254, opened 2015-01-16
Add tracking beacon to streams

commuterjoy:


Makefile
```diff
@@ -50,13 +50,13 @@ debug:
 	export ELASTIC_SEARCH_URL=${ELASTIC_SEARCH_URL}; HOSTEDGRAPHITE_APIKEY=123; export apikey=${API_KEY}; export PORT=${PORT}; node --debug-brk server/app.js
 
 build:
-	export ENVIRONMENT=development; gulp
+	export ENVIRONMENT=development; ./node_modules/.bin/gulp
```
matthew-andrews: is this not working?

commuterjoy: I don't have any globally installed node modules.

matthew-andrews: you shouldn't need to have it globally installed.  I thought node added `:node_modules/.bin:` (relative) to the path?

PR #255, opened 2015-01-16
Pagination fix

wheresrhys:


server/controllers/search.js
```diff
@@ -203,6 +203,7 @@ module.exports = function(req, res, next) {
 				mode: 'compact',
 				title: streamRequest.getHumanReadableTitle(),
 				q: streamRequest.q,
+				safeQ: encodeURIComponent(streamRequest.q),
```
matthew-andrews: ok for now, but maybe useful as a helper in the long temr?

matthew-andrews: {{encodeUriComponent}}{{q}}{{/encodeUriComponent}}

wheresrhys: good idea. I'll do it now as I need to apply the same fix in ilyushin too

PR #279, opened 2015-01-27
Output streams as an RSS feed

adgad:
@matthew-andrews @commuterjoy @wheresrhys 

screen shot 2015-01-27 at 16 52 00


commuterjoy:
You'll need to extract something from the Accept header and vary on it.

commuterjoy:
Validate it - http://validator.w3.org/feed/

commuterjoy:
With the &lt;link&gt; elements you probably don't want ever want localhost in there.

adgad:
Added vary header. I wanted to avoid hardcoding 'next.ft.com' so just taking whatever URL used to consume the feed.

matthew-andrews:
then it probably should be FT.com (with its legacy article URL schema :disappointed: )

matthew-andrews:
http://www.ft.com/cms/s/{{uuid}}.html automatically forwards to the right ID

PR #284, opened 2015-01-28
Make relative links in RSS feed go to next.ft.com

adgad:
/cc @commuterjoy


commuterjoy:
:+1: bit of a hack :)

PR #314, opened 2015-02-18
Use new init functions for ui-setup/header

keirog:
See https://github.com/Financial-Times/next-header/pull/74


adgad:
These will be breaking changes of header/ui-setup, so you'd need to bump the version numbers in bower.json - otherwise :+1:

keirog:
In case anyone is search for "what was that problem keith was having with the dobi build", if your app is moaning about a missing gulp module, you might want to run `npm cache clean`. I've added it to the Codeship setup commands for dobi but not for other apps.

PR #320, opened 2015-02-20
adding nightwatch tets

spriggsha:
Nightwatch tests to go with new provisioning steps.


commuterjoy:
:+1: assume it passes!

PR #335, opened 2015-03-02
moved featured out of stream into dobi

wheresrhys:
@adgad


wheresrhys:
see also https://github.com/Financial-Times/next-stream/pull/33 @adgad

PR #343, opened 2015-03-06
Sourcemaps

aintgoin2goa:


Makefile
```diff
@@ -58,10 +58,10 @@ debug:
 	export ELASTIC_SEARCH_URL=${ELASTIC_SEARCH_URL}; HOSTEDGRAPHITE_APIKEY=123; export apikey=${API_KEY}; export PORT=${PORT}; node --debug-brk server/app.js
 
 build:
-	export ENVIRONMENT=development; gulp
+	export ENVIRONMENT=development; ./node_modules/.bin/gulp build-dev
 
 build-production:
-	gulp
+	./node_modules/.bin/gulp build-prod
```
matthew-andrews: ./node_modules/.bin should be on your path. Shouldn't need to do this. Should just be “gulp”

matthew-andrews: gulp build-prod*

aintgoin2goa: most of the other apps are doing it this way - just trying to add some consistency as I go and this way seems more portable

PR #349, opened 2015-03-11
Git will look after your old code, if it's not used delete it… it'll only confuse people :-)

matthew-andrews:
one of the 10 commandments of assanka/ftlabs:-

- Thou shalt not commit commented out code


wheresrhys:
:+1: to this PR, but commented out code has its uses e.g. in next-sass-setup I commented out code that should work but didn't because of a bug upstream in css-clean. Bug now fixed so very easy to reinstate the code. https://github.com/Financial-Times/next-sass-setup/blob/master/main.scss#L35

> Thou shalt not commit commented out code without leaving a good reason

I'd agree with though

matthew-andrews:
ok yes there are edge cases where i also comment out & commit code, but this wasn't one of them :smile:

PR #351, opened 2015-03-12
Adding js-success test for dobi

spriggsha:


matthew-andrews:
:+1: this has already helped us fix bugs in IE10 on grumman & engels

PR #357, opened 2015-03-19
Update next-mustard to fix ios7

adgad:
/cc @matthew-andrews


matthew-andrews:
this should be meregd already

matthew-andrews:
yup it's in 2.2.2 https://github.com/Financial-Times/next-mustard/releases/tag/v2.2.2

PR #369, opened 2015-03-23
Correctly encode the Add button so it matches that of tags

adgad:
/cc @wheresrhys


adgad:
(build failing because of dodgy npm installs. master seems fine)

PR #382, opened 2015-04-07
Get rid of imagemin

matthew-andrews:


matthew-andrews:
[We no longer store binary assets in apps so it doesn't make sense for this to live here]

PR #383, opened 2015-04-13
Add no-search-results message

keirog:


keirog:
External +1 received from @wheresrhys

PR #389, opened 2015-04-16
removed index from data trackable on bundled news items

wheresrhys:


matthew-andrews:
dont think this is supported anymore

PR #390, opened 2015-04-20
Build only assets that have been modified

kaelig:
(just copied from next-grumman)


matthew-andrews:
#thought

at some point i want to centralise all these tasks into NBT
https://github.com/Financial-Times/next-build-tools/blob/master/tasks/build.js

kaelig:
Yes, but then you have a hard dependency on NBT. Meaning all apps need to always be up to date to build correctly with the latest version of NBT. Could actually be a bad-good idea.

matthew-andrews:
we have that already…  and we have lots of dupliaction and varied quality of config……  i think we've got a stable pattern now.  it'd be a good time to centralise, on balance :)

PR #391, opened 2015-04-20
Make build task phony

kaelig:
Prevents messages such as `make: 'build' is up to date.` from popping when running `make build`


matthew-andrews:
:+1:

PR #396, opened 2015-05-07
Upgrade to header 9

matthew-andrews:


matthew-andrews:
cc @ironsidevsquincy

views/layout.html
```diff
@@ -1,3 +1,22 @@
+<div class="sub-header">
+	<div class="o-grid-row">
+		<div class="sub-header__title">{{headerData.section.title}}</div>
```
adgad: Should probably be a h1?

ironsidevsquincy:
:+1:

PR #397, opened 2015-05-08
Now flags api simplified code that integrates with it can be simplified too

matthew-andrews:
cc @ironsidevsquincy


matthew-andrews:
Also remove commented out code

PR #399, opened 2015-05-08
Start to switch dobi to next-ft-api-client; enable elastic search searching for /somethingId/{base64} endpoints

matthew-andrews:


server/controllers/stream.js
```diff
@@ -9,46 +9,22 @@ var api = require('next-ft-api-client');
...
-		return article.id;
-	});
+	// HACK:MA: Remove articles that have failed to have been retrieved
+	var ids = posts.filter(function(article) { return !!article; });
 
 	return ft[getItems](ids)
 		.then(function(articles) {
 			var finalizedArticles = [];
-			// HACK:MA: Remove articles that have failed to have been retrieved
-			articles = articles.filter(function(article) {
```
matthew-andrews: this stuff should have been behind a flag I think…… going to let it “expire” and see the effect……

server/controllers/stream.js
```diff
@@ -9,46 +9,22 @@ var api = require('next-ft-api-client');
...
-				});
-			}
-
 			articles.forEach(function(article, index) {
 				if (article) {
 					if (options.isSearch) {
-						article.matchedString = new MatchHighlighter(options.q, methodeSearchArticles.filter(function(art) {
-							return art.id === article.id;
-						})[0].raw.item.summary.excerpt).run();
+						article.matchedString = new MatchHighlighter(options.q, article.raw.item.summary.excerpt).run();
```
matthew-andrews: @wheresrhys this _seems_ to achieve the same thing but much more simply……??

Makefile
```diff
@@ -14,7 +15,7 @@ test:
 	nbt verify
 	# Run all tests except for smoke tests
 	$(MAKE) build-production
-	export apikey=12345; export api2key=67890; export ELASTIC_SEARCH_URL='http://123.foundcluster.com:9200/v1_api/item'; export HOSTEDGRAPHITE_APIKEY=123; export ENVIRONMENT=production; mocha --reporter spec tests/server/
+	export apikey=12345; export api2key=67890; export ELASTIC_SEARCH_HOST='ft-elastic-search.com'; export ELASTIC_SEARCH_URL='https://ft-elastic-search.com/v1_api/item'; export HOSTEDGRAPHITE_APIKEY=123; export ENVIRONMENT=production; mocha --reporter spec tests/server/
```
ironsidevsquincy: do we still use `ELASTIC_SEARCH_URL`?

matthew-andrews: i think the old
 `ft-api-client` still uses it

PR #404, opened 2015-05-12
Redirect to id urls

matthew-andrews:
@commuterjoy @ironsidevsquincy


matthew-andrews:
wait…… search ……

matthew-andrews:
nope we're good. just my api key

ironsidevsquincy:
cool :+1:

server/controllers/redirector.js
```diff
@@ -0,0 +1,63 @@
...
+					.concat(metadata.subjects)
+					.concat(metadata.topics);
+
+				var tag = allTags.find(function(tag) {
+					return tag.term.name.toLowerCase() === thingName.toLowerCase()
+						&& tag.term.taxonomy.toLowerCase() === thingType.toLowerCase();
+				});
+
+				if (tag) {
+					res.redirect('/stream/' + thingType + 'Id/' + tag.term.id + (req.params.extension ? '.' + req.params.extension : ''));
```
commuterjoy: I wonder if we can drop the Id bit of the URL

Something like this would be a really neat URL scheme :-

 next.ft.com/articles
 next.ft.com/people
 next.ft.com/topics
 next.ft.com/organisations

matthew-andrews: yup that's the plan…… https://github.com/Financial-Times/next-letov/blob/master/server/app.js#L12-14

PR #407, opened 2015-05-13
In depths are broken anyway, so ditch the code for now…

matthew-andrews:


matthew-andrews:
this is going to need to be done much more carefully :disappointed:

PR #416, opened 2015-05-13
Add dynamic tertiary nav

i-like-robots:
Adds related navigation for pages at the bottom of the taxonomy. 

Requires:
- https://github.com/Financial-Times/next-feature-flags-api/pull/138


server/lib/related-tags.js
```diff
@@ -0,0 +1,18 @@
...
+		.then(function(response) {
+			if (response.status >= 400) {
+				throw new Error(response.status + " response from server for " + url);
+			}
+
+			return response.json();
+		})
+		.catch(function(error) {
+			throw new Error("Bad response from server for " + url);
+		});
```
ironsidevsquincy: need this catch just to throw the error again?

server/controllers/stream.js
```diff
@@ -126,31 +126,47 @@ module.exports = function(req, res, next) {
...
-						case 'personal finance':
-							partialName = 'personal-finance';
-							break;
-						default:
-							partialName = req.params.thingId;
-					}
-					stream.pathToSubSectionsPartial = 'next-navigation/templates/partials/' + partialName;
+				var currentNavItem = res.locals.nav.filter(function(item) {
+					return item.id === req.params.thingId;
+				}).pop();
```
ironsidevsquincy: this will only match top level section, but think that's ok for now

ironsidevsquincy:
:+1:

server/controllers/stream.js
i-like-robots: I presume it's OK for now as we're not currently showing sibling navigation.

PR #424, opened 2015-05-14
Headline article list

matthew-andrews:
screen shot 2015-05-14 at 09 42 50
screen shot 2015-05-14 at 09 46 08


matthew-andrews:
Ignore the fact the dates arent relative — this is a bug just affecting me locally

adgad:
Nice.

Search results should be showing the highlighted "matches" though - rather than the subheading.

matthew-andrews:
@adgad i think that's a separate regression http://next.ft.com/search?q=test

matthew-andrews:
hmm  no actually ti is a regression in this PR

PR #429, opened 2015-05-14
Rudimentary follow for core experience

adgad:
/cc @matthew-andrews @wheresrhys 

It's not a great experience - we always show 'Add' even if they've already added it, because we don't know without JS. But it's better than not working!


matthew-andrews:
:+1:

PR #437, opened 2015-05-18
Add dynamic links to subsections

i-like-robots:
- Switches out search suggestions API for slurp for relevant results
- Moves styles towards design vision

screen shot 2015-05-18 at 13 55 05


client/components/sub-header/_main.scss
```diff
@@ -58,21 +79,29 @@
 		padding-bottom: 15px;
 	}
 }
+
+.next-navigation__group__items--columns {
+	@include oGridRespondTo(M) {
+		columns: 2;
```
ironsidevsquincy: how does this look in older browsers?

client/components/sub-header/_main.scss
```diff
@@ -58,21 +79,29 @@
...
-			color: #d9d9d9;
+			color: inherit;
 			font-weight: 300;
 
 			&:hover {
 				text-decoration: underline;
 			}
 			&:after {
-				@include nextIcon(arrow-right, #d9d9d9, 10);
+				@include nextIcon(arrow-right, #f99d9d, 10);
```
ironsidevsquincy: is this is `next-colors`

ironsidevsquincy:
Awesome :+1:

client/components/sub-header/_main.scss
i-like-robots: Do we have any hooks - feature or IE specific - to write a fallback against?

client/components/sub-header/_main.scss
i-like-robots: It's defined here: https://github.com/Financial-Times/next-sass-setup/blob/master/_colors.scss#L12

but `oColorsGetPaletteColor('red-2')` has disastrous consequences :boom:

ironsidevsquincy: as in, it doesn't compile?

client/components/sub-header/_main.scss
ironsidevsquincy: hmmmm, not really. can you write it so it enhances? layout doesn't have to be perfect without `columns`

client/components/sub-header/_main.scss
i-like-robots: The function returns nothing so `nextIcon` will error out.

adgad: I had this problem just now. Moving @import next-sass-setup to the top of client/main.scss should fix it. (I think one of the origami modules was setting up o-colors without our custom defined colours)

i-like-robots: thanks @adgad :+1:

client/components/sub-header/_main.scss
```diff
@@ -82,17 +111,18 @@
...
+		padding-right: 20px;
 
-		&:nth-child(-n+3) {
-			margin-top: 0;
-		}
-		&:nth-child(3n+1) {
-			clear: left;
+		// basic columns fallback for IE9
+		float: left\9;
+		width: 50%\9;
```
i-like-robots: The best columns fallback I can think without adding feature detection @ironsidevsquincy

client/components/sub-header/_main.scss
i-like-robots: I guess the difficulty with columns is that it's specified on the parent element and any fallback would be applied to its children. I don't know of a reliable method of getting the same effect :fearful:

package.json
```diff
@@ -24,7 +24,7 @@
     "intl": "^0.1.4",
     "lodash": "^2.4.1",
     "moment": "^2.8.3",
-    "next-ft-api-client": "Financial-Times/next-ft-api-client#v3",
+    "next-ft-api-client": "Financial-Times/next-ft-api-client",
```
matthew-andrews: `^3.0.0`

matthew-andrews: [ it's published on npm now…  sorry i forgot to update ]

PR #438, opened 2015-05-18
Add logging for slurp failure

matthew-andrews:


matthew-andrews:
@i-like-robots

i-like-robots:
:ship:

PR #440, opened 2015-05-19
Remove ft-api-client; only use next-ft-api-client

matthew-andrews:
@ironsidevsquincy


server/controllers/stream.js
```diff
@@ -12,22 +11,46 @@ var relatedTags = require('../lib/related-tags');
...
+		uuid: ids,
+		useElasticSearch: options.useElasticSearch,
+		type: 'Article'
+	})
 		.then(function(articles) {
 			return {
 				articles: articles
 					.filter(identity)
 					.map(function(article) {
+						var output = { raw: article };
```
matthew-andrews: this is just to match the data models expected by next-article-card-component, etc ……… in the future this will be much simpler, honest :wink

server/controllers/stream.js
```diff
@@ -12,22 +11,46 @@ var relatedTags = require('../lib/related-tags');
 var articlePrimaryTag = require('ft-next-article-primary-tag');
 
 function processContent(posts, options) {
-	var getItems = options.useElasticSearch ? 'mget' : 'get';
 	if (!posts.length) return [];
 
 	// HACK:MA: Remove articles that have failed to have been retrieved
 	var ids = posts.filter(identity);
-	return ft[getItems](ids)
+	console.log(ids);
```
i-like-robots: =O

matthew-andrews: soz

PR #441, opened 2015-05-19
Simplify sub-header

i-like-robots:
- Use a more agnostic class name for the sub-nav header because it is copied into tools expander
- Simplify sub-header layout template logic
- Don't multi column on small screens

May be of interest to @ironsidevsquincy


matthew-andrews:
simple is good :)

PR #442, opened 2015-05-20
Make __gtg more intelligent

matthew-andrews:


matthew-andrews:
so that sensu reports ill health as frequently as this pingdom check does
https://my.pingdom.com/reports/uptime#check=1571742&tab=uptime_tab&daterange=7days

PR #445, opened 2015-05-20
Be more careful about which tags to check against

matthew-andrews:


matthew-andrews:
https://app.getsentry.com/nextftcom/ft-next-dobi/group/65620293/

PR #451, opened 2015-05-21
Try out obt beta 3

matthew-andrews:
cc @albertoelias


AlbertoElias:
Is the error due to obt?

matthew-andrews:
nope i don't think so…… not requiring any help from you here.  just wanted to let you know the trial is continuing :)

PR #452, opened 2015-05-21
Fix o-expander in sub header

i-like-robots:
Links are currently accessible but the toggle and expander are broken.


matthew-andrews:
:+1:

PR #454, opened 2015-05-21
Simplify

matthew-andrews:


server/controllers/stream.js
```diff
@@ -149,10 +108,7 @@ module.exports = function(req, res, next) {
 				res.set('Content-Type', 'text/xml');
 				res.send(streamToFeed(title, 'http://next.ft.com', streamPath, stream.items));
 			} else if (accepts === 'html') {
-
-				stream.currentNavItem = res.locals.nav.find(function(item) {
-					return item.id === req.params.thingId;
-				});
+				stream.currentNavItem = res.locals.nav.find(function(item) { return item.id === req.params.thingId; });
```
i-like-robots: Maybe assign the `children` property to `currentNavItem` here to simplify the template.

i-like-robots:
I definitely prefer the article transformation living in its own module.

matthew-andrews:
own module or own repo?

i-like-robots:
I was complimenting your choice of code reorganisation! I think the lack of emoji made the comment ambiguous.

This was doesn't: :+1:

matthew-andrews:
emoji is important!  ok thanks i've implemented your suggestion, i think

i-like-robots:
Exactly that, it's .1% nicer now =)

PR #461, opened 2015-05-22
Not all orgs have industry classifications

matthew-andrews:


matthew-andrews:
https://app.getsentry.com/nextftcom/ft-next-dobi/group/66410705/

PR #467, opened 2015-05-22
Update in the news nav to be correct on semantic headers

matthew-andrews:
does this look ok to you  @i-like-robots  ?


i-like-robots:
All good :ship:

PR #471, opened 2015-05-26
Remove profiles from organisation pages :'(

matthew-andrews:


commuterjoy:
If you want to bring them back, stick part of the summary (eg. the long bit) behind a toggle then count active usage. It would be nice to say, 'summaries were used by 40% of people' or whatever.

PR #473, opened 2015-05-26
deal with cases where no exact tag match is found

matthew-andrews:
@ironsidevsquincy


matthew-andrews:
e.g. http://next.ft.com/stream/organisations/Samsung

ironsidevsquincy:
hmmm, seems a bit dangerous. might get back to the old 'lex' => 'alexander' etc issue

matthew-andrews:
it's only used as an emergency backup for when exact matches aren't there……
e.g. http://next.ft.com/stream/organisations/Samsung

i can make the detection a bit clever to deal with the lex->alexander possibility

ironsidevsquincy:
there will always be edge cases, lex -> lex luthor

what's still linking to the old section urls? can we fix those?

matthew-andrews:
this particular example is next-mustang i think.  @wheresrhys , do you have any idea why “Samsung” appears in search results when it doesn't exist as a metadata concept as far as CAPI is concerned? :/

wheresrhys:
A task for MUM. My guess is it's changed from Samsung to Samsung Ltd or something in methode but not been updated in SAPI.

matthew-andrews:
@wheresrhys i think it has been changed to:-
Samsung Electronics Co Ltd
api.ft.com/content/items/v1/fcba0c7c-5aa7-390a-8f77-492378af2e6b

screen shot 2015-05-26 at 13 21 48

Rhys, would you be able to drop a note to contentplatform.support@ft.com describing in what way that tag exists in SAPI and request it to be changed to ‘Samsung Electronics Co Ltd’…… it's not very clear to me how this exists in the Search API…… I thought for a tag to be in the Search API it need it needed to be tagged against some content………??

matthew-andrews:
thanks to @wheresrhys we've isolated this bug in the search api

other affected examples include:-
- IBM seems to be now ’International Business Machines Corp’, breaking this: http://next.ft.com/stream/organisations/IBM
- Amazon is now ‘Amazon.Com Inc’, breaking this:- http://next.ft.com/stream/organisations/Amazon
- Ted Cruz should be Teddy Cruz - breaking this:- http://next.ft.com/stream/people/Ted%20Cruz

PR #474, opened 2015-05-26
Migrate related tags functionality from Slurp to ES

i-like-robots:
So we're not relying on an unsupported API =)

Requires https://github.com/Financial-Times/next-ft-api-client/pull/30

Before:
screen shot 2015-05-26 at 17 12 44

After:
screen shot 2015-05-26 at 17 12 31


matthew-andrews:
undefined is not a function :)

i-like-robots:
Depends on https://github.com/Financial-Times/next-ft-api-client/pull/30 being merged and bumped

i-like-robots:
@matthew-andrews Should be good to go

matthew-andrews:
:+1:

PR #478, opened 2015-05-27
Support pages, e.g. indepths in Next

matthew-andrews:
screen shot 2015-05-27 at 12 45 59

cc @wheresrhys


server/controllers/pages.js
```diff
@@ -0,0 +1,34 @@
...
+
+var api = require('next-ft-api-client');
+var identity = require('../lib/identity');
+var articleModelV1 = require('../models/article-v1');
+var onBadServerResponse404 = require('../lib/on-bad-server-response-404');
+var renderStream = require('../lib/render-stream');
+
+module.exports = function(req, res, next) {
+	api.pages({
+		uuid: req.params.thingId
```
wheresrhys: how're you getting the id and how are redirects from ft.com/indepth/blah-de-blah gonna work?

matthew-andrews: next-redirector gonna provide that magic

matthew-andrews: also planning to move this stuff https://github.com/Financial-Times/next-dobi/blob/master/server/controllers/redirector.js into next-redirector too

PR #479, opened 2015-05-27
allow interspersing other items in the article list

wheresrhys:
@matthew-andrews we're planning on adding some followy stuff midstream. As it may be contentious I wanted to add this little featurette in a separate PR


matthew-andrews:
hmm…… can we chat about this first

matthew-andrews:
:+1:

PR #481, opened 2015-05-28
Add markets data quote to organisation streams

i-like-robots:
Uses the Markets Data API (via our proxy) to fetch the latest quote for an organisation and displays it in the sub-header. Based on the new markets pages (E.G. https://markets.ft.com/data/equities/tearsheet/summary?s=us:AAPL)

screen shot 2015-05-28 at 11 02 25

We should monitor this to see how it affects our TTFB.


commuterjoy:
Does this peg our performance or stability to a third party?

i-like-robots:
We're hitting our proxy of the markets data API so we can tweak our levels of caching etc.

https://github.com/Financial-Times/next-dobi/commit/ee04c6d70c10642d72dcf5042a1469eecea8acf4#diff-9997625e6106b308242004626edae0e1R24

Like the related navigation this should fail quietly and continue to output the page.

server/lib/markets-data.js
```diff
@@ -0,0 +1,51 @@
...
+			return response;
+		})
+		.then(fetchres.json)
+		.then(function(response) {
+			if (response.data.items && response.data.items.length) {
+				return formatMarketsData(response.data.items.pop());
+			}
+		})
+		.catch(function(error) {
+			logger.error(error);
```
matthew-andrews:
```js
if (!(error instanceof fetchres.BadServerResponse)) {
  raven.captureError(error);
}
```?

ironsidevsquincy: personal opinion, but I feel sentry should be for unexpected errors, things we didn't account for that bubbled up to the top. If we think there's a chance something might go wrong, e.g. with have a `catch`, then log it into splunk. Makes parsing easier, and doesn't muddy genuine errors.

Also `raven.captureError` doesn't give you any useful information, like url it occurred on, client details, etc.

matthew-andrews: i think this would catch unexpected errors (hence the `if` that prevents catching the bad server/network errors) e.g. something going wrong in `formatMarketsData`.

You're right that it won't capture useful request/client info……

ironsidevsquincy: yeah, but we're still continuing with the execution so it's semi-expected, an informative error really. 

sentry's good for immediate feedback that something's completely broken the site, cluttering it with general errors dilutes the useful noise.

PR #488, opened 2015-06-02
Organisation sub header improvements

i-like-robots:
screen shot 2015-06-02 at 16 47 27

screen shot 2015-06-02 at 16 52 47


matthew-andrews:
:+1:

views/partials/sub-header.html
```diff
@@ -3,11 +3,8 @@ <h1 class="sub-header__page-title">{{title}}</h1>
 	{{#if conceptId}}
 		{{>next-myft-ui/templates/follow light=true}}
 	{{/if}}
-	{{#if symbol}}
-		<a class="sub-header__symbol" href="https://markets.ft.com/data/equities/tearsheet/summary?s={{symbol}}" data-trackable="markets-link">{{symbol}}</a>
-	{{/if}}
 </div>
-{{#if currentNav}}
+{{#if currentNav.length}}
```
matthew-andrews: actually wait

matthew-andrews: is it possible to fix this in the controller so that we simply don't set this property if there is no data?

matthew-andrews: [just so it's consistent with `dynamicTertiaryNav`]

i-like-robots: Haha I thought this might trigger a reaction.  `getCurrentNav()` will always give us an array as is, I guess we can decide whether to not to add the `currentNav` property to the stream based on that.

i-like-robots: @matthew-andrews updated

i-like-robots:
This will also close #487

matthew-andrews:
:+1:

PR #491, opened 2015-06-03
Don't retry on mapping and organisation endpoints when they fail (so …

matthew-andrews:
…that we fail faster and show the critical content when those services encounter difficulties)

cc @i-like-robots (also https://github.com/Financial-Times/next-ft-api-client/pull/32)


matthew-andrews:
currently causing this:- next.ft.com/stream/organisationsId/TnN0ZWluX09OX09yZ2FuaXNhdGlvbnNfVVNfMjAwOTA5MTZfMjYy-T04=

server/controllers/v1-taxonomy.js
```diff
@@ -79,12 +79,12 @@ module.exports = function(req, res, next) {
...
 
 			if (stream.taxonomy === 'organisations' && res.locals.flags.semanticHeader) {
-				promises.push(api.mapping(stream.tag.term.id, stream.tag.term.taxonomy)
+				promises.push(api.mapping(stream.tag.term.id, stream.tag.term.taxonomy, { retry: 0 })
 					.then(function(mapping) {
 						if (!mapping) {
 							throw new Error('Mapping doesn‘t exist');
 						}
-						return api.organisations({ uuid: mapping.apiUrl.replace(/^http:\/\/api\.ft\.com\/organisations\//, '') });
+						return api.organisations({ uuid: mapping.apiUrl.replace(/^http:\/\/api\.ft\.com\/organisations\//, ''), retries: 0 });
```
i-like-robots: Using both `retry` and `retries` causes a bit of a twitch.

matthew-andrews: oops that's a typo.

so annoyed with myself.  should have use `retries` in the first place.  but now sued `retry` better keep it for consistency :(

PR #493, opened 2015-06-03
Re-scope sub-header navigation styles

### LINKED ISSUE #492: header display issues when components missing
screen shot 2015-06-03 at 11 45 33

cc @i-like-robots

i-like-robots:
Fixes #492 

screen shot 2015-06-03 at 12 07 17


ironsidevsquincy:
could event use `next-dobi-` as a prefix...

ironsidevsquincy:
could even use `next-dobi-` as a prefix...

i-like-robots:
Everything for now is `.sub-header` ... which could potentially conflict with a more generic sub-header component I guess. But for now we'll live.

matthew-andrews:
^^ @ironsidevsquincy @i-like-robots planning to rename this app to `next-stream` at some point :)

PR #494, opened 2015-06-03
Hackily bring back markets data

matthew-andrews:
[will be improved when we can use concordances api curerntly in progress in capi and we wont worry about the missing enriched goodness]


matthew-andrews:
cc @i-like-robots

i-like-robots:
I don't think I even want to know why https://github.com/Financial-Times/next-ft-api-client/commit/e1be5208d9d736431b6a121e3625e8bebe3aa44e was necessary! I look forward to the concordances API.

matthew-andrews:
because the organisation's sematnic goodness has disappeared (i've asked @railsagainstignorance but no response yet)……  i don't think we need to worry about it…… real concordance api coming soon and we can leave dobi on that branch and create a new, fresh `.concordance` method of next-ft-api-client until then……

PR #498, opened 2015-06-04
Use the subheading rather than body

matthew-andrews:


matthew-andrews:
fixes this…… https://app.getsentry.com/nextftcom/ft-next-dobi/group/67850700/

PR #499, opened 2015-06-04
Add icons to org summaries, refactor some sub header class names

i-like-robots:
screen shot 2015-06-05 at 16 12 45

Requires https://github.com/Financial-Times/next-sass-setup/pull/52


wheresrhys:
Could we commission similar icons for topics and regions too, then we could use them at the foot of articles as well. They look far slicker than the existing origami ones

matthew-andrews:
:+1: sorry about the merge cornflakes

PR #508, opened 2015-06-05
fix the thing rhys broke then cheekily raised as a bug

matthew-andrews:
!isTaxonomySearch replaced with isFreeTextSearch


matthew-andrews:
cc @wh

matthew-andrews:
@wheresrhys

PR #517, opened 2015-06-08
Do not clone sub sections nav into tools with JS

i-like-robots:
Fixes https://github.com/Financial-Times/next-dobi/issues/496

Should also fix https://app.getsentry.com/nextftcom/ft-next-frontend/group/67731344/ (I think caused by having two elements on page with same class being targeted by `document.querySelector`)


matthew-andrews:
:+1: gzip is your friend

PR #518, opened 2015-06-10
Add related video and MPU

ironsidevsquincy:
screen shot 2015-06-10 at 11 37 46


matthew-andrews:
woww!!

views/partials/article-list.html
```diff
@@ -1,15 +1,32 @@
...
+		{{#slice items limit="7"}}
+			<li class="stream-item stream-item--row o-grid-row">
+				{{#if template}}
+					{{{usePartial template}}}
+				{{else}}
+					{{>next-card/templates/card}}
+				{{/if}}
+			</li>
+		{{/slice}}
+		{{#if showRelatedContent}}
```
matthew-andrews: hmm seems like we have two solutions here…… could we not just put the related content thing in the stream array itself with the  `template` set to not an article card?

matthew-andrews: [instead of slicing in the view]

ironsidevsquincy: yep, good point

server/controllers/v1-taxonomy.js
```diff
@@ -62,7 +62,10 @@ module.exports = function(req, res, next) {
 				layout: 'article-list',
 				symbol: symbol,
 				currentNav: currentNav && currentNav.children ? currentNav : false,
-				taxonomy: taxonomy
+				taxonomy: taxonomy,
+				// only show on top level sections for now
+				showRelatedContent: res.locals.flags.relatedContentOnStreams &&
```
matthew-andrews: also should we limit this to just be on page 1 of streams?

ironsidevsquincy: hmmm...for now, yes

client/components/stream/related.js
```diff
@@ -0,0 +1,79 @@
...
+		if (!relatedItemsContainerEl) {
+			return;
+		}
+		var tagName = relatedItemsContainerEl.getAttribute('data-tag-name');
+		var searchTagName = tagName === 'Life & Arts' ? 'Life and Arts' : tagName;
+		var searchQuery = 'all=tag:section:' + searchTagName;
+
+		fetch('//next-brightcove-proxy.ft.com/search?' + searchQuery + '&page_size=2')
+			.then(function (response) {
+				return response.json();
```
matthew-andrews: what's the behaviour if this response is not `.ok`?

client/components/stream/related.js
```diff
@@ -0,0 +1,79 @@
...
+			})
+			.then(function (json) {
+				if (json.items.length === 0) {
+					return;
+				}
+				var relatedItemsTitleEl = document.createElement('h2');
+				relatedItemsTitleEl.className = 'related-items-title';
+				relatedItemsTitleEl.innerHTML = '<span class="related-items-title__tag-name">' + tagName +
```
matthew-andrews: es6 template string might be neater here, if you like :-)

matthew-andrews: [very very optional]

ironsidevsquincy: Amazing. Is that browserify doing the magic?

matthew-andrews: OBT :wink:

ironsidevsquincy: ahh, uses babelify, need to make more use of this

matthew-andrews: client side only atm

PR #522, opened 2015-06-11
Live stream MVP

i-like-robots:
live-stream


client/components/stream/_main.scss
```diff
@@ -126,3 +98,35 @@
...
+	@include nextButtonsBase(
+		$background: oColorsGetPaletteColor('red-1'),
+		$color: oColorsGetPaletteColor('white')
+	);
+	width: 100%;
+	margin: 10px 0;
+}
+
+.live-stream__log {
+	// .visuallyhidden from HTML5BP
```
phamann: Do we have any utility classes? i.e. this could be abstracted to `u-visually-hidden` for us to use globally.

adgad:
Not to speak for design, but my guess is the button shouldn't be red, since that's reserved for myFT.... @simonwilliamsFT @pauloneillft

client/components/stream/lib/live-stream.js
```diff
@@ -0,0 +1,127 @@
...
+		this.streamItems = this.stream.querySelectorAll('[data-content-id]');
+
+		this._lastUpdateHTML = null;
+
+		this.options.onUpdate && this.options.onUpdate(this._lastUpdateID);
+	}
+
+	createUpdateButton() {
+		let button = document.createElement('button');
+		button.setAttribute('aria-controls', this.target.id);
```
phamann: Nice! :sunglasses:

client/components/stream/_main.scss
wheresrhys: or `[aria-hidden] {display: none}`

Similarly with js we could do with a few utilties

i-like-robots: Not ATM. I'm all in favour of adding `.u-` prefixed utils too but I think there would be some debate about where they should live.

i-like-robots:
Good point, I shall seek design eyes.

client/components/stream/lib/live-stream.js
```diff
@@ -0,0 +1,127 @@
...
+		this.streamItems = this.stream.querySelectorAll('[data-content-id]');
+
+		this._lastUpdateID = this.streamItems[0].dataset.contentId;
+
+		this.logArea = this.createLogArea();
+		this.updateButton = this.createUpdateButton();
+
+		this.target.appendChild(this.logArea);
+
+		this.poll();
```
phamann: Is this feature wrapped in a switch? if not could you create one to disable polling, as could be used to lighten load on origin servers.

client/components/stream/lib/live-stream.js
```diff
@@ -0,0 +1,127 @@
...
+	// Poll rate in milliseconds
+	updateRate: 1000 * 10,
+	// Class name/s for update button
+	buttonClassName: 'live-stream__update',
+	// Content for update button
+	buttonInnerHTML: 'View new articles',
+	// Class name/s for log area
+	logAreaClassName: 'live-stream__log',
+	// Log message when new articles are available
+	logMessageActive: 'New articles are available, click to view them',
```
i-like-robots: As a side note, telling a screen reader user to click the button is probably not smart. We'd have to agree some global keyboard shortcuts.

simonwilliamsFT:
red isn't reserved for myFT , red is reserved for topics (people / places /
events )  themselves, clicking a red word should take you to a page of
content related to that word ,  it  has been used for follow buttons,
primarily because it's prominence means you can get more clicks to drive
follow engagement and they should generally appear next to topics, although
even there weight may want to come down eventually.


On 12 June 2015 at 12:17, Arjun Gadhia <notifications@github.com> wrote:

> Not to speak for design, but my guess is the button shouldn't be red,
> since that's reserved for myFT.... @simonwilliamsFT
> <https://github.com/simonwilliamsFT> @pauloneillft
> <https://github.com/pauloneillft>
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/Financial-Times/next-dobi/pull/522#issuecomment-111456303>
> .
>



-- 

*Simon Williams*
Design Manager, FT.com <http://www.ft.com/>
One Southwark Bridge, London SE1 9HL
Phone: 020 7873 4323

-- 

------------------------------
This email was sent by a company owned by Pearson plc, registered office at 
80 Strand, London WC2R 0RL.  Registered in England and Wales with company 
number 53723.

client/components/stream/lib/live-stream.js
i-like-robots: I don't think I follow, what sort of switch?

I'm presuming we'll get a 304 99% of the time from the cache, otherwise a user's data will take the hit too.

matthew-andrews: :+1: for safety switch

i-like-robots: I'll add one now... what am I adding? A toggle in the UI?

phamann: Reason being: if origin server is 50x'ing you don't want to continue to hammer it with requests, turning feature off may allow the service to heal itself e.t.c and prevents [dog-piling](https://en.wikipedia.org/wiki/Cache_stampede)

client/components/stream/lib/live-stream.js
```diff
@@ -0,0 +1,127 @@
...
+
+	poll() {
+		this._poll && clearTimeout(this._poll);
+		this._poll = setTimeout(this.update.bind(this), this.options.updateRate);
+	}
+
+	update() {
+		this._poll = null;
+
+		fetch(this.targetURL)
```
phamann: Timeout?

client/components/stream/lib/live-stream.js
matthew-andrews: a feature flag?

matthew-andrews: http://github.com/Financial-Times/next-feature-flags-api

i-like-robots: @matthew-andrews Feature flag is on the back end, it appends a class and an ID in the template
@phamann I'll count errors and switch off polling when it reaches a threshold

client/components/stream/lib/live-stream.js
matthew-andrews: caution: client side fetch doesn't support timeout

matthew-andrews: https://github.com/whatwg/fetch/issues/20

client/components/stream/lib/live-stream.js
i-like-robots: adding a `maxErrors` option in c3a8e982c0550def5b349c45960cfc818e6d478a which will terminate the poll after the given number of consecutive errors.

matthew-andrews:
:+1: ?

i-like-robots:
Didn't merge on Friday due to continued Browserstack IE10 fails making me look bad.

PR #523, opened 2015-06-12
using concept ids and tidying up search suggestions

wheresrhys:
@i-like-robots


bower.json
```diff
@@ -7,7 +7,8 @@
     "next-sass-setup": "^6.0.0",
     "o-expander": "^1.0.0",
     "o-buttons": "^2.0.3",
-    "next-card": "^5.2.0"
+    "next-card": "^5.2.0",
+    "fetchres": "~1.1.0"
```
matthew-andrews: i don't think necessary to add this here when it's already there in `package.json`

wheresrhys: It'll work without it , yes, but I don't think we should blur the lines on how we manage dependencies client side.

matthew-andrews: i think you'll get away with it here because I think browserify will pick one and only use one but because of the `instanceOf` fetchres simply _doesn't work_ if JavaScript picks one in one place and another in another place……

matthew-andrews: much much safer to just have it once

client/components/sub-header/main.js
```diff
@@ -1,25 +1,61 @@
...
+				var html = '<h3 class="sub-header__section__heading">Did you mean</h3><ul class="sub-header__navigation-group sub-header__navigation-group--columns">';
+				var totalChars = 0;
+				tags.forEach(function (tag) {
+					var url = tag.url;
+					if (!url) {
+						url = tag.id ? ('/stream/' + tag.taxonomy + 'Id/' + tag.id) : ('/stream/' + tag.taxonomy + '/' + tag.name);
+					}
+					html += '<li class="sub-header__navigation-group__item"><a class="sub-header__navigation-group__link" data-trackable="link" href="' + url + '" title="' + tag.taxonomy + ': ' + tag.name + '">' + tag.name + '</a></li>';
+				});
+				html += '</ul></div>';
```
i-like-robots: Could this be formatted more nicely as a template string if we can ES6 it up on the client side?

client/components/sub-header/main.js
```diff
@@ -1,25 +1,61 @@
...
+				document.querySelector('.sub-header__section').insertAdjacentHTML('afterend', '<div class="sub-header__section js-search-suggestions" data-o-grid-colspan="hide M4 L5 Moffset4 Loffset3" data-trackable="search-suggestions">' + html);
```
i-like-robots: Duplicating the column layout in JS could be confusing later, maybe this could always live in the template as it should collapse when empty?

PR #530, opened 2015-06-15
Ensure fetch sends cookies

i-like-robots:


matthew-andrews:
ah yes. credentials.  not cors.  still pleased i managed to guess it!

i-like-robots:
There's also the `mode` option for which `same-origin` is a valid value but it doesn't _appear_ as though that's also required - not that I can make much sense of the docs TBH.

matthew-andrews:
:+1:

PR #531, opened 2015-06-15
IE10

matthew-andrews:


Makefile
```diff
@@ -2,7 +2,7 @@ API_KEY := $(shell cat ~/.ftapi 2>/dev/null)
 API2_KEY := $(shell cat ~/.ftapi_v2 2>/dev/null)
 GIT_HASH := $(shell git rev-parse --short HEAD)
 TEST_HOST := "ft-next-dobi-branch-${GIT_HASH}"
-TEST_URL := "http://ft-next-dobi-branch-${GIT_HASH}.herokuapp.com/"
+TEST_URL := "ft-next-dobi-branch-${GIT_HASH}.herokuapp.com"
```
matthew-andrews: i'm also going to…… at some point…… tidy up these env vars…… the host is not a host and this is no longer a URL and it was always not the TEST URL…………

i-like-robots:
Definitely needs more commits! If CI passes both times then let's do it.

i-like-robots:
Keen.

matthew-andrews:
soz

matthew-andrews:
i recreated something suspiciously like the bug we saw

matthew-andrews:
and fixed it

matthew-andrews:
if it happens again just add the `--env` option to the `nightwatch` test and i'll look into it again :innocent:

PR #533, opened 2015-06-15
most styling and handling of placeholder for video now done in n-video

wheresrhys:


client/components/stream/_related.scss
```diff
@@ -59,97 +68,16 @@
...
-		left: 50%;
-		margin-top: -20px;
-		margin-left: -20px;
-		transition: opacity 0.1s;
-	}
-	&:hover:after {
-		opacity: 0.5;
-	}
-}
-.related-item__content {
```
ironsidevsquincy: reckon these should stay in here, quite specific

PR #540, opened 2015-06-16
If the systems are healthy don't ever show stale content

matthew-andrews:


i-like-robots:
Yep, we're meant to be fast moving after all! :+1:

PR #542, opened 2015-06-16
IE9 can't talk x-domain

ironsidevsquincy:
So handle these errors, i.e. stops https://app.getsentry.com/nextftcom/ft-next-frontend/group/65544907/


matthew-andrews:
are we saying we don't support those features in IE9?  I suspect we ought to…… maybe we need to think about jsonp

ironsidevsquincy:
agreed we ought to. good point, our old friend jsonp, will add it to the brightcove api endpoint

matthew-andrews:
for now this is ok though :)

PR #547, opened 2015-06-17
Use JSONP if needed

ironsidevsquincy:


client/components/stream/related.js
```diff
@@ -22,7 +23,8 @@ module.exports = {
 		var streamId = location.pathname.split('/').pop();
 		var tagName = relatedItemsContainerEl.getAttribute('data-tag-name');
 
-		return fetch('//next-video.ft.com/stream/' + streamId + '?page_size=6')
+		var fetchFn = ('XDomainRequest' in window) ? nJsonpFetch : fetch;
+		return fetchFn('//next-video.ft.com/stream/' + streamId + '?page_size=6')
```
matthew-andrews: this `if` could even go in the `n-jsonp-fetch`

eg.
```js
let nJsonpFetch = require('n-jsonp-fetch/automatic');

nJsonpFetch('//next-video.ft.com/stream/' + streamId + '?page_size=6');
```

just a thought

ironsidevsquincy: yeah, agreed, have done this in two separate places now, not sure it should live in `n-jsonp-fetch` though, feel like it might be polluting it slightly, e.g. sometimes does jsonp, sometimes doesn't, dependent on the client. maybe another super-thin wrapper that just does this

PR #550, opened 2015-06-18
Engaged myFT follow call-to-action

phamann:
This introduces a new semi-intrusive myFT follow call-to-action for engaged users of stream pages. If a user has visited a given stream page more than three times within a week and are *not* currently following the tag, we display a fixed prompt to suggest them to follow.

- Queries the myFT client to assert whether they are following or not
- Queries the myFT user history API to assert whether they are a frequent visitor
- On mobile injects CTA into stream in the 6th item position
- On desktop displays CTA in a fixed oOverlay
- Gives the user to dismiss all CTAs like this in the future and persists their decision to localStorage

We will eventually A/B test this along side another less intrusive CTA being built within the next couple of days.

engaged-follow

/cc @adgad @wheresrhys


client/components/engaged-follow/_main.scss
```diff
@@ -0,0 +1,82 @@
+@import "o-overlay/main";
+@import "next-type/main";
+
+// scss-lint:disable ImporantRule
```
adgad: typo in important?

adgad:
LGTM :+1:

PR #556, opened 2015-06-18
Use concordance API

matthew-andrews:


matthew-andrews:
Don't merge! Need to request our API keys all get access to the concordance API.

PR #564, opened 2015-06-24
Hub pages

adgad:
/cc @matthew-andrews @i-like-robots 

MVP for grouping streams by topic. Behind a switched off flag - not limited to particular streams yet, because it's quite interesting to see it at work on other streams.

screen shot 2015-06-24 at 16 22 02


matthew-andrews:
cool :+1:

PR #565, opened 2015-06-24
404 on v1 taxonomy pages that have no content

matthew-andrews:
So I can track them in splunk and fix them in harrier ^^

cc @adgad


adgad:
:+1:

PR #568, opened 2015-06-25
Sort streams by "trending"

adgad:
/cc @matthew-andrews @commuterjoy 
Unstable and behind an 'off' flag - this sorts streams based on a weighting of recency/popularity.


commuterjoy:
nice

matthew-andrews:
cool

PR #572, opened 2015-06-26
Only init live streams on page 1, fixes #562

### LINKED ISSUE #562: Auto-refreshing of content should probably be disabled on page 2+
Otherwise the old content that you want to see gets overwritten by the new content :-)

@i-like-robots

i-like-robots:


ironsidevsquincy:
:+1:

PR #573, opened 2015-06-26
Show ratings on cards

adgad:
/cc @matthew-andrews


server/models/article-v1.js
```diff
@@ -25,6 +25,13 @@ module.exports = function(article, opts) {
 		output.showMedia = true;
 	}
 
+	if(article.nextHackRating) {
```
commuterjoy: hacks should have expiry dates!

matthew-andrews: this would be hard to expire :grin:

matthew-andrews: ah we could have a flag for the rating ^^

matthew-andrews:
awesomes

server/app.js
```diff
@@ -21,6 +21,7 @@ var app = module.exports = express({
 
 app.get('/special-reports.?:extension(rss)?', require('./controllers/special-reports'));
 app.get('/news-feed.?:extension(rss)?', require('./controllers/news-feed'));
+app.get('/top-rated/:thingId', require('./controllers/top-reviews'));
```
matthew-andrews: might want to add the taxonomy in here too…… `/:thingType/:thingId` so we can do, e.g. top rated in London

http://next.ft.com/streams/regionsId/TnN0ZWluX0dMX0dCX0VOR19Mb25kb24%3D-R0w%3D

PR #577, opened 2015-06-30
Related video errors

tom-parker:
cc @phamann @ironsidevsquincy


ironsidevsquincy:
probably only want to do this for 404s, i.e. if the message is `Brightcove proxy responded with 404`

tom-parker:
True, should we be reporting brightcove proxy errors from within another app though? E.g if the proxy is down would we want to know about it here?

matthew-andrews:
adam b fixed ths in https://github.com/Financial-Times/next-dobi/pull/601

PR #600, opened 2015-07-09
Brightcove: Only throw an error if it's not a 4xx response

adambraimbridge:


client/components/stream/related-video-stream.js
```diff
@@ -30,7 +30,13 @@ var exec = function () {
 				if (response.ok) {
 					return response.json();
 				} else {
-					throw new Error(`Brightcove proxy responded with ${response.status}`);
+
+					// Only throw an error if it's not a 4xx response
+					if (response.status.toString()[0] !== "4") {
```
matthew-andrews: ah i think this check should be done within the `catch` block rather than here…… i think it's right to throw an error at this point……

client/components/stream/related.js
```diff
@@ -23,10 +23,19 @@ module.exports = {
 					return response.json();
 				} else {
 					relatedItemsStreamItemEl.parentNode.removeChild(relatedItemsStreamItemEl);
-					throw new Error(`Brightcove proxy responded with ${response.status}`);
+
+					// Only throw an error if it's not a 4xx response
+					if (response.status.toString()[0] !== "4") {
+						throw new Error(`Brightcove proxy responded with ${response.status}`);
```
matthew-andrews: same here…… you won't need to do the `if (!json || !json.items) {` check if you throw here

next-team:
I've spotted a [difference](https://s3-eu-west-1.amazonaws.com/ft-next-qa/image_diffs/dobi/2015-07-09/12:10-2199/failures/index.html) between your pull request and production — [see full output](https://s3-eu-west-1.amazonaws.com/ft-next-qa/image_diffs/dobi/2015-07-09/12:10-2199/successes/index.html)

client/components/stream/related-video-stream.js
adambraimbridge: Thanks Matt. I'm trying to understand this better ... 
(1) In this file, the `catch` completely ignores the error, so isn't it pointless to throw the error at all? 
(2) In the other file (related.js), we still want to pay attention to non-4xx errors, yeah?

adambraimbridge:
Closing this PR in lieu of https://github.com/Financial-Times/next-dobi/pull/601

PR #627, opened 2015-07-28
Special reports articles have to be in a specialReport primarySection

ironsidevsquincy:
e.g. https://next.ft.com/50f209f4-3494-11e5-b05b-b01debd57852 is tagged with `specialReport`, so is appearing in the Aerospace special report, but is not actually part of the official special report

Now, only consider articles to be part of a special report if their `primarySection` is a `specialReport`

cc @matthew-andrews


matthew-andrews:
¯\_(ツ)_/¯ ok then

PR #635, opened 2015-07-30
use id not name to filter photo diary entries from world stream

andygnewman:
@matthew-andrews 
One that was missed.


matthew-andrews:
:+1:

PR #638, opened 2015-07-30
time -> brand-and-time

wheresrhys:
@keirog I can't test this locally as having weird wifi problems and can't run any app, but if you can find an instance of this on a page might be worth  amore general once over to make sure it works with new cards


keirog:
:-1: down with card-with-comment.html

PR #652, opened 2015-08-05
hotter hot items

i-like-robots:
Use flexbox so each 'card' best fits its container without a bunch of breakpoints repeated for each combination of columns

screen shot 2015-08-06 at 11 52 03
screen shot 2015-08-06 at 11 52 56
ie8

- [x] Chrome 
- [x] Firefox
- [x] IE10
- [x] IE8-9


views/partials/special-report-item.html
```diff
@@ -1,40 +1,36 @@
 <div class="hot-topics__topic hot-topics__topic--special-report" data-trackable="stream-special-report">
 	{{#ifEquals @root.featuredItems.length 1}}
 		{{#if image}}
-			<div class="hot-topics__topic__background" style="background-image: url('{{#resize 1440}}{{image.url}}{{/resize}}')"></div>
+			<!-- <div class="hot-topics__topic__background" style="background-image: url('{{#resize 1440}}{{image.url}}{{/resize}}')"></div> -->
```
ironsidevsquincy: delete?

ironsidevsquincy:
:+1:

matthew-andrews:
holy moly those fonts look awful

i-like-robots:
It's one of the lesser concerns in IE8 TBH. If we want to do core experience we could build in IE8/9 first!

PR #654, opened 2015-08-06
Fix live stream updates after html change

ironsidevsquincy:
Use class hook to get the node

cc: @i-like-robots


i-like-robots:
:+1:

PR #657, opened 2015-08-06
Use private cache

ironsidevsquincy:


matthew-andrews:
private for everything?

ironsidevsquincy:
any reason not to? possibly json requests don't need it, but might do ab tests on them down the line, so probably easiest just to have a blanket rule

matthew-andrews:
in that case we could do it in next-express?

matthew-andrews:
:+1:

ironsidevsquincy:
Yeah, would make sense if it's a global thing. We'll discuss tomorrow

PR #659, opened 2015-08-06
add data-trackable to hot comment links

andygnewman:
so can track usage in beacon


views/partials/hot-comment-item.html
```diff
@@ -1,4 +1,4 @@
-<a href="/{{id}}" class="hot-comment__link">
+<a href="/{{id}}" class="hot-comment__link" data-trackable="hot-comment-link-{{variation}}">
```
i-like-robots: Maybe this attribute should stick to being `hot-comment-link` and then the variation described with the `data-trackable-meta` attr?

andygnewman: This is where the behaviour of data-trackable-meta is defined;

https://github.com/Financial-Times/next-beacon-component/blob/84e3580756173699f1dd32e5e1b3f4d40d9e6618/src/analytics/cta.js

Think best to keep as suggested in code above.

i-like-robots:
:white_check_mark:

PR #662, opened 2015-08-07
Add variations to hot topics to handle cases with no images

i-like-robots:
screen shot 2015-08-07 at 10 34 59
screen shot 2015-08-07 at 10 35 05


andygnewman:
+1

PR #669, opened 2015-08-11
502 when search api unavailable

matthew-andrews:


matthew-andrews:
@aintgoin2goa mind giving this a quick once over?

PR #670, opened 2015-08-11
Remove comment in stream code

adgad:
/cc @matthew-andrews @i-like-robots @andygnewman


i-like-robots:
Apologies for calling another new feature 'Hot Comments'! Looks good :+1:

PR #675, opened 2015-08-13
Ads update

robin-marr-ft:
Adding a leaderboard ad

cc/ @keirog @matthew-andrews


wheresrhys:
:+1: I wonder when/if we should be thinking about doing a similar thing for myft streams?

PR #678, opened 2015-08-17
Delete streamTopicGroups and streamTrendingSort code

adgad:
/cc @ironsidevsquincy @i-like-robots @andygnewman


i-like-robots:
:ship:

PR #681, opened 2015-08-20
Added n-message-prompts client

ifyio:
cc/ @commuterjoy @tom-parker 

This is the initial MVP of the message prompts.


commuterjoy:
+1

PR #684, opened 2015-08-26
hot topics design unification

andygnewman:
screen shot 2015-08-26 at 11 46 27
screen shot 2015-08-26 at 11 46 43


client/components/stream/_hot-topics.scss
```diff
@@ -73,30 +68,21 @@
...
-	position: absolute;
-	top: 0;
-	right: 0;
-	bottom: 0;
-	left: 0;
-	padding: 10px;
-	color: white;
+	@include nTypeBravo(4);
+	white-space: nowrap;
+	width: calc(100% -16px);
```
ironsidevsquincy: how does this look in IE<9?

client/components/stream/_hot-topics.scss
```diff
@@ -73,30 +68,21 @@
...
+	overflow: hidden;
+	text-overflow: ellipsis;
+	text-transform: uppercase;
+	text-align: center;
+	padding: 8px;
+	color: oColorsGetPaletteColor('white');
 
 	// IE8-9 don't display the image
-	background-color: oColorsGetPaletteColor('red-1');
+	background-color: oColorsGetPaletteColor('claret');
```
adgad: could this be `oColorsGetColorFor(topic, text)`

ironsidevsquincy: and there's also an alias for `oColorsGetPaletteColor`- https://github.com/Financial-Times/next-sass-setup/blob/master/_functions.scss#L5

adgad: Yes, I can add that to next-sass-setup now

andygnewman: @adgad 
As it's a claret background in this instance calling getColorFor(topic, text) would be confusing.

adgad: Hmmm...maybe we could have a topic-block use case? with background claret and text=white...

client/components/stream/_hot-topics.scss
andygnewman: @ironsidevsquincy 
in IE9 - no image
screen shot 2015-08-26 at 15 36 05

this is how it currently looks on live in IE9
screen shot 2015-08-26 at 15 40 46

ironsidevsquincy: what about 8? I more just wondering, as `calc` is unavailable for <9, what's the consequences visually?

andygnewman: Managed to get my IE8 VM working, looked OK, but I did revisit the need for calc and found I didn't need it, so good challenge.

client/components/stream/_hot-topics.scss
```diff
@@ -73,30 +68,20 @@
...
-	right: 0;
-	bottom: 0;
-	left: 0;
-	padding: 10px;
-	color: white;
+	@include nTypeBravo(4);
+	white-space: nowrap;
+	overflow: hidden;
+	text-overflow: ellipsis;
+	text-transform: uppercase;
```
i-like-robots: We get in trouble every time we do this so beware!

client/components/stream/_hot-topics.scss
```diff
@@ -114,28 +99,27 @@
...
+	@include nTypeCharlie(4);
+	@include nLinksHeadline('article-comment');
+	margin: 0;
 }
 
 .hot-topics__list-item-date {
-	color: oColorsGetPaletteColor('cold-1');
+	@include nTypeAlpha(5);
+	text-transform: uppercase;
+	color: getColorFor(timestamp, text);
```
i-like-robots: other strings are quoted so let's keep it consistent

client/components/stream/_hot-topics.scss
andygnewman: Ok, will await the wrath of editorial, but they get the win of not having the title over the image.

client/components/stream/_hot-topics.scss
andygnewman: OK will change it here (and in the one from the original implementation of hot topics that I copied the syntax from)

PR #689, opened 2015-08-28
design unification special reports hot topics

andygnewman:
/cc @ironsidevsquincy 

Transform special reports in hot topics from this;

screen shot 2015-08-28 at 12 30 35

To this;

screen shot 2015-08-28 at 12 30 44


client/components/stream/_hot-topics.scss
```diff
@@ -111,6 +111,17 @@
 	color: getColorFor('timestamp', 'text');
 }
 
+.hot-topics__special-report__tag {
+	@include nTypeAlpha(5);
+	position: absolute;
+		bottom: 0;
+		right: 0;
```
ironsidevsquincy: indentation

views/partials/special-report-item.html
```diff
@@ -16,6 +16,8 @@
 			<ol class="hot-topics__list">
 				{{#each articles}}
 					<li class="hot-topics__list-item hot-topics__list-item--{{@index}}" data-trackable="article">
+						<a class="hot-topics__list-item-link" href="/{{id}}" data-trackable="link">{{headline}}</a>
+						<br/>
```
ironsidevsquincy: can put a `display: inline-block` on the above `a` instead of this `br`

server/lib/stream-special-reports.js
```diff
@@ -1,6 +1,7 @@
 'use strict';
 
 var articleModelV1 = require('../models/article-v1');
+var extractImage = require('./extract-image');
```
ironsidevsquincy: is this new?

andygnewman: a new bit of code for extracting image from article used in hot topics, comment, analysis and now special reports.

PR #692, opened 2015-09-02
Added bower resolution for `next-js-setup`

ifyio:
This is so that builds would no longer fail


ifyio:
Closing this as the issue seems to have been resolved

PR #699, opened 2015-09-16
Stale cache handling

commuterjoy:
Instructs the cache to update the content every 2 minutes but if the origin is down then show stale content for 3 days.

https://docs.fastly.com/guides/performance-tuning/serving-stale-content


server/lib/cache-control.js
```diff
@@ -1,5 +1,5 @@
 'use strict';
 
 module.exports = {
-	cache: { 'Surrogate-Control': 'max-age=60' }
+	cache: { 'Surrogate-Control': 'max-age=60,stale-while-revalidate,stale-if-error=259200' }
```
matthew-andrews: think `stale-while-revalidate` needs an `=something`

PR #702, opened 2015-09-18
None of that old API client stuff thank you

i-like-robots:
OMG so last year


matthew-andrews:
:+1:

matthew-andrews:
wow how do you put up with these tests …… they fail so often ……

i-like-robots:
This is a legit fail

i-like-robots:
Ah, lies. It's flakey.

PR #713, opened 2015-09-24
Search through all articles for a special report image

ironsidevsquincy:
### Before

screen shot 2015-09-24 at 11 59 40

### After

screen shot 2015-09-24 at 11 59 44


server/lib/stream-special-reports.js
```diff
@@ -43,7 +43,9 @@ module.exports = function (items) {
 				});
 				if (articles.length) {
 					var specialReport = articles[0] && articles[0].raw.item.metadata.primarySection.term;
-					var specialReportImage = extractImage(articles[0]);
+					var specialReportImage;
+					// find an article with an image
+					articles.some((article) => specialReportImage = extractImage(article));
```
andygnewman: You might also want to give it a better default size, eg.

    extractImage(article, 540)

https://github.com/Financial-Times/next-stream-page/blob/master/server/lib/extract-image.js

ironsidevsquincy: Looks like the `n-topic` card doesn't use the picturefill, it's hardcoded to 300px

https://github.com/Financial-Times/n-topic/blob/master/templates/card.html#L9

PR #716, opened 2015-10-02
Keep page titles consistent

i-like-robots:
screen shot 2015-10-02 at 11 04 45

Switch stream page title to Financier for consistency with article page.


i-like-robots:
Design says no, keep Financier for headlines only.

PR #740, opened 2015-10-26
reinstate non-es for text search and bump next-ft-api back up

andygnewman:
reinstate non-ES article search for keyword search
bump next-ft-api up to 3.15.0

cc @matthew-andrews


matthew-andrews:
:+1:

PR #741, opened 2015-10-27
Add local nightwatch json

leggsimon:


Makefile
```diff
@@ -52,4 +52,4 @@ clean-deploy: clean install build-production deploy
 
 smoke:
 	nbt test-urls ${TEST_HOST}
-	export TEST_URL=${TEST_URL}; nbt nightwatch test/browser/tests/*
+	export TEST_URL=${TEST_URL}; nbt nightwatch test/browser/tests/* --config ./nightwatch.json --env chromeLatest,firefoxLatest,ie11,ie10,ie9
```
matthew-andrews: what's the latest story on ios/android?

leggsimon: They are still being fixed, Tanya will let me know when they should be working.

matthew-andrews:
Sorry to be a pedant but would you mind tidying up the whitespace in the `nightwatch.json` file … i think we normally indent json with 2 spaces but it doesn't really matter as long as it's consistent :-)

matthew-andrews:
Otherwise :+1:

PR #745, opened 2015-11-02
Fix resolutions

matthew-andrews:


keirog:
Dupe of https://github.com/Financial-Times/next-stream-page/pull/744

PR #746, opened 2015-11-03
Fix sauce labs notify

ironsidevsquincy:


matthew-andrews:
stream page doesn't use saucelabs anymore… it uses the ft grid… so this code can be deletedd actually :)

ironsidevsquincy:
Oh right. Is that something we're rolling out to all apps? Where can I see more info on this ft grid - usage, where is it, etc?

matthew-andrews:
@ironsidevsquincy not yet, just trialling…  if we make our repos private we won't be able to use sauce labs anymore (unless we pay… which we decided, as an ft, not to do) so i wanted to understand whether we **could** use the internal grid to understand the benefits/disadvantages so that we can decide to move everything there, or overturn the saucelabs decision

matthew-andrews:
details are here: https://sites.google.com/a/ft.com/technology/system/app/pages/customSearch?scope=search-ns&q=ft+grid

matthew-andrews:
also worth saying we haven't decided to make repos private yet, but i wanted to understand what work we would have to do if we needed to do that

ironsidevsquincy:
Ok. Nightwatch is still being run as part of stream's build - should this be turned off? And where are the build step are we running ft grid?

matthew-andrews:
`nightwatch` runs on FT's Grid.

matthew-andrews:
this is the command that runs nightwatch on the FT grid: https://github.com/Financial-Times/next-stream-page/blob/master/Makefile#L55

PR #754, opened 2015-11-20
Use V1 labels on Organisation and Stream pages

matthew-andrews:
@andygnewman


matthew-andrews:
Fixes, for example, 习近平 => Xi Jinping

PR #756, opened 2015-11-23
Start a React-ion of the stream page … sigh

matthew-andrews:
Justification:-

- Stream pages (e.g. FastFT) will have more advanced auto-reloading functionality which will not be user triggered, so accessibility type things like keeping selected text is more important than before.

Note the React version of the stream page is just a simple list for now…


components/stream-items/index.js
```diff
@@ -0,0 +1,36 @@
...
+		const items = context.items.map(item => {
+				return <li>{item.headline}</li>;
+			});
+
+		return (
+			<section role="section">
+				<div className="o-grid-row">
+					<div data-o-grid-colspan="full-width">
+						<h2 className="stream__title">{title}</h2>
+						<ol className="stream__list js-stream-list" data-trackable="{context.trackable}" id="{odId}">{items}</ol>
```
ironsidevsquincy: minor: no need for quotes around attribute values if the values in `{...}`

`data-trackable="{context.trackable}" => data-trackable={context.trackable}`

components/stream-items/index.js
```diff
@@ -0,0 +1,36 @@
...
+		const sectionClass = (context.isLive ? 'live-stream js-live-stream' : '') + ' ' + (context.isHubPage ? 'topic-stream' : '');
+		const odId = context.isLive ? 'live-stream-items' : '';
+
+		const title = context.isSpecialReport
+			? `There are ${context.indexCount} articles in this special report`
+			: `Latest ${context.titlePreposition} ${context.title}`;
+
+		const relatedVideo = this.props.flags.relatedVideoStream
+			? <ol class="related-video-stream o-grid-row" data-trackable="related-video-stream"></ol>
+			: '';
```
ironsidevsquincy: *think* it's better to use `null`, rather than empty string

components/stream-items/index.js
```diff
@@ -0,0 +1,36 @@
+import React, {Component} from 'react';
+
+export default class Layout extends Component {
+
+	render () {
+		const context = this.props.context;
+
+		const sectionClass = (context.isLive ? 'live-stream js-live-stream' : '') + ' ' + (context.isHubPage ? 'topic-stream' : '');
```
ironsidevsquincy: where's this used?

server/app.js
```diff
@@ -1,43 +1,13 @@
...
+		'es6.arrowFunctions',
+		'es6.destructuring',
+		'es6.modules',
+		'es6.parameters',
+		'es6.spread',
+
+		'react',
+		'strict'
+	]
+});
```
ironsidevsquincy: is this needed for react?

ironsidevsquincy:
Worth a test, :+1: 

One thing I'm struggling with (though quite possibly due to my limited React knowledge) is how to change the HTML outside of react, i.e. we use o-date, n-video, etc, to update on the client, but React has a minor freak-out when we do this, along the lines of 'something else has update the dom which wasn't me, so I'm reverting it back to what I know'

server/app.js
matthew-andrews: just copy and pasted from front page…… definitely need babel for JSX

components/stream-items/index.js
matthew-andrews: ok

components/stream-items/index.js
matthew-andrews: fixed now

server/app.js
ironsidevsquincy: have a card to test the performance of using babel register. guess it's ok to add to streams for now

matthew-andrews: let me see if i can eliminate it

matthew-andrews:
I've reduce the babel-usage down to just JSX

PR #761, opened 2015-11-25
Fast FT page to V3

matthew-andrews:
Realise this data structure will soon change but this turned out to be the easiest starting place and other streams can build on this as they go from V1/V2 => V3


matthew-andrews:
This will also add pagination for Fast FT (will now allow usage to reach more than the previous 50 posts) 

:smile:

PR #762, opened 2015-11-25
Convert news-feed to V3

matthew-andrews:


matthew-andrews:
I am starting to think our various different data formats, centralised helper methods, different versions of dependencies and inconsistent applications of the different approaches to the above make things more complicated than just duplicating the implementations across the apps.

@i-like-robots @andygnewman

PR #766, opened 2015-11-25
V3 org people

matthew-andrews:


server/controllers/v1-v2-taxonomy.js
```diff
@@ -61,71 +62,68 @@ module.exports = function(req, res, next) {
...
 			var organisation = data[1];
-			var searchQuery = req.params.thingType + 'Id:"' + tmeId + '"';
-
-			return api.searchLegacy({
-					query: searchQuery,
-					offset: perPage.offset,
-					count: perPage.count,
-					useElasticSearch: res.locals.flags.elasticSearchItemGet,
-					useElasticSearchOnAws: res.locals.flags.elasticSearchOnAws,
-					includePodcasts: res.locals.flags.podcasts
```
matthew-andrews: probably need to add this back here and in `news-feed` :grin:

matthew-andrews: https://github.com/Financial-Times/next-ft-api-client/blob/e8d6219c2c266220c227daea5012c500f050d310/src/search-legacy.js#L35

PR #771, opened 2015-11-30
When no summary, default to empty string

matthew-andrews:
to fix this:-
screen shot 2015-11-30 at 11 05 08


server/models/article-v3.js
```diff
@@ -25,7 +25,8 @@ module.exports = function(article) {
 		showSubheading: true,
 		subheading: article.summaries && article.summaries[0]
 			// HACK: For Fast FT, when it moves to WordPress shouldn't need this anymore
-			|| article.annotations.some(tag => tag.prefLabel === 'fastFT') && $('p').first().text(),
+			|| article.annotations.some(tag => tag.prefLabel === 'fastFT') && $('p').first().text()
+			|| '',
```
andygnewman: Shouldn't we put in a 'falsy' value eg. undefined / null and handle in the template?

matthew-andrews:
Yup you're right,  updated this PR to do that

andygnewman:
:+1:

PR #774, opened 2015-12-01
Upgrade special reports to V3

matthew-andrews:


server/controllers/special-reports.js
```diff
@@ -9,58 +9,92 @@ const signedFetch = require('signed-aws-es-fetch');
...
+										aggs: {
+											id: {
+												terms: {
+													field: 'idV1',
+													size: 1
+												}
+											},
+											taxonomy: {
+												terms: {
+													field: 'taxonomy',
```
i-like-robots: I guess this is unnecessary as we know this already, I only included it to check it was working =]

PR #777, opened 2015-12-01
Add @charypar style animations to the lively updating pages

matthew-andrews:


charypar:
Actually @imranolas style animations :)

PR #782, opened 2015-12-02
Get the el fresh every time

matthew-andrews:


matthew-andrews:
there's a better way to do this

PR #788, opened 2015-12-02
Restrict scope to React to just the stream list

matthew-andrews:


views/topics.html
```diff
@@ -1,37 +0,0 @@
-{{>sub-header}}
```
matthew-andrews: this was completely unused

matthew-andrews:
Also reduces the amount of data transferred from `.json` endpoints

PR #796, opened 2015-12-04
image -> mainImage

matthew-andrews:


matthew-andrews:
Need to fix missing images on related block of special reports streams

PR #801, opened 2015-12-04
Prepare to move fields

matthew-andrews:


matthew-andrews:
By removing these from the fixtures (and checking they aren't used by the helpers) we can confidently move where these fields are computed from after coming from ES to just before being passed to `render-stream`.

PR #805, opened 2015-12-04
Less output in json

matthew-andrews:


server/lib/streamToFeed.js
```diff
@@ -12,9 +12,8 @@ module.exports = function(title, domain, streamPath, items) {
 		items.forEach(function(item) {
 			if(item && item.id) {
 				feed.item({
-					title: item.headline,
-					url: domain + '/' + item.id,
-					author: item.authors && item.authors.length ? item.authors[0].name : '',
+					title: item.title,
+					url: domain + '/content/' + item.id,
```
matthew-andrews: @i-like-robots found one more old reference to the article URLs!!!!!

matthew-andrews: ^^ the !!!'s were because i was amazed how many we have

matthew-andrews:
This is ready to go :grin: @andygnewman

matthew-andrews:
125kb -> 23kb :smile:

Issue #810, opened 2015-12-07
[open] button in stream needs tracking


matthew-andrews:

andygnewman:
Done and works
https://beacon.ft.com/explorer?query%5Bevent_collection%5D=cta&query%5Banalysis_type%5D=count&query%5Bgroup_by%5D=meta.domPath&query%5Btimezone%5D=UTC&query%5Bfilters%5D%5B0%5D%5Bproperty_name%5D=meta.domPath&query%5Bfilters%5D%5B0%5D%5Boperator%5D=in&query%5Bfilters%5D%5B0%5D%5Bproperty_value%5D%5B0%5D=stream%20%7C%20article-open&query%5Bfilters%5D%5B0%5D%5Bproperty_value%5D%5B1%5D=stream%20%7C%20article-close&query%5Btimeframe%5D=this_1_days

PR #813, opened 2015-12-08
add data trackable to in stream articles and open close button

andygnewman:
So can track article actions as being part of stream.


matthew-andrews:
:+1:

matthew-andrews:
fixes #810

PR #817, opened 2015-12-08
[WIP] refactor react card for stream pages

andygnewman:
- [x] use string interpolation
- [x] refactor open / close button logic
- [x] abstract out inline article react class from stream card
- [x] enable appending to dom using bare fragment option from article & styling
- [x] flag for iframe vs fetch
- [x] get o-share bower component working server side
- [x] get all usual article functions working (eg. share, print, follow links in article)


matthew-andrews:
ha, nice, you're quite right.  probably a lot can be tidied up in this case.

matthew-andrews:
@andygnewman I think this works well enough now — printing (albeit a bit weird), sharing, saving…

server/controllers/v2-taxonomy.js
```diff
@@ -31,7 +31,8 @@ module.exports = function(req, res, next) {
 				items: results[0]
 					.map(articleModel({
 							hasActions: config.hasActions,
-							hasOpen: config.hasOpenInStream
+							hasOpen: config.hasOpenInStream,
+							useDomInsertion: res.locals.flags.streamOpenArticleDOM && isFastFt
 						})),
```
andygnewman: replace isFastFT with config.hasOpenInStream

andygnewman:
The print one will look a bit weird - probably need to do a scope on the print function to just print the article elements of the DOM.

matthew-andrews:
Hmm…  I think this is also missing galleries and videos…

andygnewman:
Ah yes, galleries get stuck on through client side, are you seeing the placeholders come through?

matthew-andrews:
No, but I'm wondering whether it would be better to take out the duplicated CSS/JS from article for now and bring it back when we figure how we're going to share the code.

andygnewman:
Yes, would be a good idea, as it is only for internal proof of concept at the moment.

PR #819, opened 2015-12-09
ADS-598 inserted leaderboard ad slot in stream page headers

leggsimon:
@andygnewman


andygnewman:
:+1:  though bizarre you have to put in into 3 partials, but that's our issue to solve...

leggsimon:
@andygnewman yeah I thought that was strange, I just put them back in the ones I removed them from.

PR #824, opened 2015-12-14
Cunning plan

matthew-andrews:


andygnewman:
Looks OK. When I was trying it, I wasn't sure at what point the share.init() goes to work vs the dom object it's looking for gets rendered, ie. if the init happens before the elements are in the dom, does it miss the opportunity to do it's job?

matthew-andrews:
good point, i've updated it now ^^

PR #828, opened 2015-12-17
accept-encoding already set in fastly

wheresrhys:
@matthew-andrews


matthew-andrews:
:+1:

matthew-andrews:
thanks for getting to the bottom of this.  i admit i was just blindly putting this here

Issue #534, opened 2015-06-15
Live streaming sometimes streams in old content


matthew-andrews:
cc @i-like-robots 

![wat](https://cloud.githubusercontent.com/assets/825088/8169310/a4beb682-13a0-11e5-9602-4e90bad1b41d.gif)

So what I think is happening here is different cache nodes have different versions of the page………

Kind of funny :smile:

i-like-robots:
Surely we're not caching pages for that length of time!? I have two ideas of how to work around this:

1. Include the time of the initial/most recent page update in the query string (may cause loads of cache misses)
2. Test if the last story ID in the response header is on the page already

matthew-andrews:
but if we're on page 2 of a paginated feed we shoudl never have live content stream in, should we?  is that even useful?

i-like-robots:
Different issue https://github.com/Financial-Times/next-dobi/issues/562

matthew-andrews:
oops

matthew-andrews:
totally got confused here

PR #833, opened 2016-01-05
WIP Dynamic loading & notification of new stories (FastFT)

andygnewman:
highlight

scroll

- Live streams fetch data every 60 seconds
- New stories are inserted into stream if the top of stream is in view...
- Or the user is alerted to the presence of new stories


wheresrhys:
Are we doing anything to purge the fastft url on fastly when a new fastft story is published to es?

i-like-robots:
@wheresrhys No... we did this last time https://github.com/Financial-Times/next-stream-page/issues/534

i-like-robots:
Actually that was a different issue...

matthew-andrews:
\o/

PR #834, opened 2016-01-07
Remove broken tests

commuterjoy:
These don't work in production, so seems unfair to include them as smoke tests.

Master branch has been [failing for a week](https://circleci.com/gh/Financial-Times/next-stream-page) because of this.


matthew-andrews:
i thought @i-like-robots was fixing these

matthew-andrews:
:+1: for now, but we should fix them

commuterjoy:
Yup, I couldn't actually see them being used anywhere, but I didn't look very hard.

matthew-andrews:
they're pulled in on the front end…

matthew-andrews:
https://github.com/Financial-Times/next-stream-page/blob/09c541f5e397ba122b370a17007306269e98e785/client/components/stream/special-report.js#L5

i-like-robots:
It's on the list, will prioritise

PR #835, opened 2016-01-08
Fix special reports related content

i-like-robots:
- Migrates related tags from ES v1/CAPI v1 to ES v3
- Refactors special reports fetching into shared module so it can also be used by related special reports.

## known issues

- Related special reports are unstyled. Looks like this was missed when hot topic cards were migrated to use n-topic component.


matthew-andrews:
:+1:

PR #841, opened 2016-01-13
Fixes the gap for hot-comments with no image

keirog:
Also:
- removes the `hot-comment__story--no-image` modifier, in favour of no-image by default
- makes `hot-comment__story--has-image` easier to search for

Before:
before
After:
after


laurieboyes:
super :+1:

PR #842, opened 2016-01-14
when doing hot topics etc get twice as many stories from ES

andygnewman:
Hot Topics, Hot Opinion, Hot Analysis are generated from the items pulled from ES.
Our standard get from ES is 50 items.
This gets 100 from ES when we are doing Hotness...but still limits the displayed story list to 50.
In this way we can more consistently fulfill the maximum quota of Hotness.


matthew-andrews:
sorry again for breaking this

server/lib/search-es.js
```diff
@@ -19,7 +19,7 @@ module.exports = (opts) => {
 	return signedFetch('https://next-elastic.ft.com/v3_api_v2/_search', {
 			method: 'POST',
 			body: JSON.stringify({
-				size: opts.perPage.count,
+				size: opts.perPage.initialCount || opts.perPage.count,
```
matthew-andrews: it's probably safer to put this option on the `opts` object rather than on the `perPage` object.

Issue #848, opened 2016-01-15
relatedTaxonomy end point doesn't fail gracefully


andygnewman:
The relatedTaxonomy end point (eg. /special-reports/MDEzZjdiOTAtYTFhNC00NzBjLTk2ZWMtMWFlOTZmZTM5Yzkz-U3BlY2lhbFJlcG9ydHM=/people)
Has not failed gracefully when no stories in the list match the taxonomy (returns a 500)

https://github.com/Financial-Times/next-stream-page/pull/835 introduced a rewrite, may need some kind of error handling. 

cc: @i-like-robots

PR #849, opened 2016-01-15
change smoke test url

andygnewman:
Change of smoke test url as unwanted failure behaviour on existing one.

See this issue https://github.com/Financial-Times/next-stream-page/issues/848


matthew-andrews:
:+1:

PR #856, opened 2016-01-19
initial testing setup for react components using jsdom and sinon

i-like-robots:
- Implements new spec runner for React components (`make component-specs`)
- Uses jsdom to mock browser environment on the server
- Adds a helper to use Sinon to stub/spy on React class methods and mock the clock
- Includes a basic spec for the top-level Stream component.

screen shot 2016-01-19 at 16 13 52

Considered and tried using Jest which does the jsdom and dependency mocking for you but couldn't get it to work.

Probably worth looking further at Shallow rendering, possibly using a tool such as https://github.com/glenjamin/skin-deep

CC @adgad @andygnewman


andygnewman:
neat

PR #857, opened 2016-01-19
Open article in stream - redirect to article page if access denied

andygnewman:
If user is denied access to the article opened in stream, redirect to the article page where the proper barrier will be rendered.


matthew-andrews:
:+1:

PR #858, opened 2016-01-19
Refactor taxonomy controller

callumlocke:
Here's my first attempt at refactoring the v1-taxonomy controller...

- upgraded to Node 5 and Babel 6
- changed some things to be more ES6
- rewrote the controller using async/await, to get rid of the then-then-then noise and make it generally more readable imo
  - if using ES2016 is too controversial, it would be easy to change this to a wrapped generator (which is what Babel compiles async functions into anyway), replacing the `await` statements with `yield`, and it would work exactly the same, just a tiny bit less pretty.

---

EDIT:
- this PR no longer upgrades Node and Babel
- the controller no longer uses async/await, no more linting issues


callumlocke:
Ah this fails at `nbt verify` because of the async function. This could be fixed by [updating nbt](https://github.com/Financial-Times/next-build-tools/issues/329), or just by switching to the generator approach I mentioned above.

It also fails at `nbt provision` with this error (not sure why):

```
Must run `nbt build` (or `nbt hash-assets`) before running `nbt deploy-hashed-assets`
```

matthew-andrews:
maybe to unblock this keep the `then` `then` `then` `catch` for now and tackle that as a separate (smaller) refactor once we have the requisite changes to nbt merged in?

callumlocke:
lessons:

- upgrading to Node 5 didn't work; we'll need to do this upgrade carefully in its own PR if at all.
  - things seem to work fine, but ultimately the main.js and main.css assets are 404s when deployed.
  - the build [seems to go fine](https://circleci.com/gh/Financial-Times/next-stream-page/1293) until the saucelab tests, which fail due to the lack of javascript.
  - unclear if this is a problem with obt being incompatible with node 5 – matt tested obt itself in node 5 and [it passed](https://travis-ci.org/Financial-Times/origami-build-tools/builds/103624336).
  - the public files **do** get created correctly when running `make build-production` locally.
  - one other slim possibility is that, in node 5 on prod, the public folder just isn't being served up by express, e.g. due to being created with bad permissions.

- upgrading to Babel 6 didn't work.
  - OBT uses a fixed (Babel 5) version of babel-loader with webpack, and this fails if it encounters Babel 6 style config.
  - tried making `client/.babelrc` to give special Babel v5 style config for the clientside build, but oddly babel-loader still complains at invalid javascript coming from inside the `server` folder – not sure why webpack would be looking in this folder at all. (unless something in server/lib is being isomorphically required by the client js?)

ironsidevsquincy:
re: babel 6 for server side, specify the plugins in the `app.js` file to avoid webpack conflicts - we're using on the front page https://github.com/Financial-Times/next-front-page/blob/master/server/app.js

callumlocke:
@ironsidevsquincy thanks will try that in a separate PR.

package.json
```diff
@@ -14,6 +14,7 @@
   },
   "dependencies": {
     "babel": "^5.8.34",
+    "bluebird": "^3.1.1",
```
i-like-robots: Native promises?

matthew-andrews: we've used co before…
https://github.com/Financial-Times/next-build-tools/blob/master/package.json#L12

i don't know if there's advantages / disadvantages to using bluebird's coroutines over that

matthew-andrews: https://github.com/tj/co

server/controllers/v1-taxonomy.js
```diff
@@ -1,273 +1,252 @@
...
-function keyValueArrayToObject(array) {
-	var object = {};
-	array.forEach(function(item) {
+function keyValueArrayToObject (array) {
+	const object = {};
+	array.forEach(item => {
 		object[item.key] = item.value;
 	});
 	return object;
 }
```
i-like-robots: Array reduce is good for this too (not saying change it, just handy):

```
array.reduce((map, item) => (map[item.key] = item.value) && map, {});
```

package.json
callumlocke: @i-like-robots Not using it to replace Promise, just using its coroutine utility.

@matthew-andrews I considered co for this, but went with bluebird because I'm anticipating making use of at least one other of its utility methods for the brexit stuff. (Also bluebird's performance is amazing -  see `promises-bluebird-generator.js` vs `generators-tj-co.js` in this [benchmark](https://github.com/petkaantonov/bluebird/tree/master/benchmark).)

server/controllers/v1-taxonomy.js
callumlocke: that's neater, although wouldn't it fail if an item's value is falsey? the `&&` wouldn't work in that case. I suppose a comma would do it but that's probably getting a bit tricksy

i-like-robots: Jaja, just thinking out loud!

callumlocke:
Can I go ahead and merge this?

PR #860, opened 2016-01-20
Open article in stream - barrier redirect, error notification, specs and styling

andygnewman:
- [x] when opening article invokes barrier, direct reader to article page, stop rendering article fragment
- [x] better separation of concerns between react components
- [x] improved CSS naming
- [x] notify the user if unable to load the requested article
- [x] specs for the open in article features

Further to do in another PR

- logging of warnings / errors (upgrade next-ft-logger to latest version across stream app)
- tracking of article open as an article view


client/components/card-open-in-stream/_main.scss
```diff
@@ -76,41 +76,36 @@
...
 
-.card-openable--article-open__wrapper {
+.card-openable__article-open-wrapper {
 	text-align: right;
 	margin-bottom: 16px;
 }
 
-.card-openable--is-open {
-	.card-openable--subheading {
+.inline-article--is-open {
```
i-like-robots: where does `inline-article` come from?

components/card-open-in-stream/inline-article/index.js
```diff
@@ -2,17 +2,26 @@
 
 const React = require('react');
 
+
 module.exports = React.createClass({
 
 	getInitialState () {
-		return { opened: 'closed', linkLabel: 'Open', inlineArticle: null, scrollY: null};
+		return {
+			opened: 'is-closed',
```
i-like-robots: Huh?

components/card-open-in-stream/inline-article/index.js
```diff
@@ -2,17 +2,26 @@
...
-		if (this.state.opened === 'closed') {
-			this.props.getCardHeight();
-			this.setState({opened: 'loading', linkLabel: 'Loading'});
+		if (this.state.opened === 'is-closed') {
+			this.setState({
+				rectInitialHeight: this.refs.container.getBoundingClientRect().height,
+				opened: 'is-loading',
+				linkLabel: 'Loading'
+			});
 			fetch(`/content/${this.props.item.id}?fragment=true&bare=true`, { credentials: 'same-origin' })
```
i-like-robots: Might be good to move the fetch into its own method so it's easier to mock out in the spec

components/card-open-in-stream/inline-article/index.js
```diff
@@ -20,26 +29,63 @@ module.exports = React.createClass({
...
 			})
 			.then(html => this.articleLoaded(html));
-		} else if (this.state.opened === 'open') {
-			this.setState({opened: 'closed', linkLabel: 'Open', inlineArticle: null});
-			this.props.articleStateChanged('closed');
-			this._article.focus();
+		} else if (this.state.opened === 'is-open') {
+			this.articleClosing();
 		} else {
 			// do nothing it's loading
```
i-like-robots: Tell assistive device users something is happening?

client/components/card-open-in-stream/_main.scss
andygnewman: Have renamed this to be consistent

client/components/card-open-in-stream/_main.scss
```diff
@@ -1,116 +1,116 @@
...
 	margin-bottom: 16px;
+	display: none;
+	.enhanced & {
+		display: block;
+	}
 }
 
-.card-openable--is-open {
-	.card-openable--subheading {
+.card-openable__inline-article--is-open {
```
i-like-robots: IMO state (usually from JS) and variations should be separate which would make `.card-openable__inline-article.is-open` - what do you think?

client/components/inline-article/_main.scss
```diff
@@ -1,4 +1,4 @@
-//new styles
+//new styles need to correct the BEM naming style when do n-article
```
i-like-robots: TODO

components/card-open-in-stream/inline-article/index.js
```diff
@@ -48,17 +102,34 @@ module.exports = React.createClass({
...
+					tabIndex="-1"
+					aria-expanded={`${this.state.articleExpanded}`}
+					className={`${this.state.articleExpanded ? 'card-openable__inline-article--is-open' : ''}`}>
+				<p className="card-openable__subheading">{this.props.item.subheading}</p>
+				{inlineArticle}
+				<div className="card-openable__article-open-wrapper">
+					<button
+						ref="button"
+						aria-controls={`id-${this.props.item.id}`}
+						aria-busy={`${this.state.articleLoading}`}
```
i-like-robots: `aria-busy={this.state.articleLoading}` is fine

components/card-open-in-stream/inline-article/index.js
```diff
@@ -48,17 +102,34 @@ module.exports = React.createClass({
...
+						className="card-openable__article-open-button"
+						onClick={this.handleClick}
+						data-trackable={`article-${this.state.buttonLabel}`}>
+						{`${this.state.buttonLabel}`}
```
i-like-robots: Same here =]

client/components/card-open-in-stream/_main.scss
andygnewman: agree

test/components/inline-article.test.js
```diff
@@ -0,0 +1,101 @@
...
+		it('when article received it is rendered correctly and button changes text', () => {
+			instance.articleLoaded(html);
+			expect(() => TestUtils.findRenderedDOMComponentWithClass(instance, 'card-openable__article')).to.not.throw();
+			expect(instance.refs.content.childNodes[1].firstChild.className).to.contain('dummy-article');
+			expect(instance.refs.button.textContent).to.equal('Show less');
+		});
+
+		xit('when article is not received an error should be rendered', () => {
+			instance.articleNotLoaded();
+			// what should be expected??
```
i-like-robots: Could check the state is reset (Button text should be 'Show more', state should be not busy) and if you're stubbing n-notification we could check that's been called (or event triggered).

test/components/inline-article.test.js
```diff
@@ -0,0 +1,101 @@
...
+
+		it('should request the article when Show more', () => {
+			expect(button.textContent).to.equal('Show more');
+			expect(instance.state.articleExpanded).to.be.false;
+			TestUtils.Simulate.click(button);
+			expect(fetchArticle.calledOnce).to.be.true;
+		});
+
+		it('should remove the article when Show less and change button label', () => {
+			instance.setState({articleExpanded: true});
```
i-like-robots: Does this throw a deprecation warning?

andygnewman: No

components/card-open-in-stream/inline-article/index.js
```diff
@@ -2,44 +2,101 @@
...
-				}
-			})
-			.then(html => this.articleLoaded(html));
-		} else if (this.state.opened === 'open') {
-			this.setState({opened: 'closed', linkLabel: 'Open', inlineArticle: null});
-			this.props.articleStateChanged('closed');
-			this._article.focus();
+	handleClick () {
+		if (!this.state.articleExpanded) {
+			this.refs.button.disabled = true;
```
i-like-robots: this should be handled with state

components/card-open-in-stream/inline-article/index.js
```diff
@@ -2,44 +2,101 @@
...
+		.then(response => {
+			if (response.ok) {
+				const accessDecision = response.headers.get('FT-Access-Decision');
+				if (accessDecision === 'DENIED') {
+					window.location.replace(`/content/${this.props.item.id}`);
+				} else {
+					return response.text().then(this.articleLoaded);
+				}
+			} else {
+				this.articleNotLoaded();
```
i-like-robots: throw the error (if you can work it out) and call `articleNotLoaded` in catch block

components/card-open-in-stream/inline-article/index.js
```diff
@@ -2,44 +2,101 @@
...
-		this.setState({opened: 'open', linkLabel: 'Close', inlineArticle: html});
-		this.props.articleStateChanged();
-		this._article.focus();
+		this.setState({
+			articleExpanded: true,
+			articleLoading: false,
+			buttonLabel: 'Show less',
+			inlineArticle: html
+		});
+		this.refs.button.disabled = false;
```
i-like-robots: As above

components/card-open-in-stream/inline-article/index.js
```diff
@@ -2,44 +2,101 @@
...
+		window.scrollTo(0, originalPosition);
+	},
+
+	articleNotLoaded () {
+		this.errorNotification();
+		this.setState({
+			articleLoading: false,
+			buttonLabel: 'Show more'
+		});
+		this.refs.button.disabled = false;
```
i-like-robots: And here

components/card-open-in-stream/inline-article/index.js
i-like-robots: Something like 

```
return response.text().then(text => throw(text))
```

PR #863, opened 2016-01-21
Don't map content publishedDate to datePublished.

i-like-robots:
No need to do this anymore!

Also fixes this issue with special reports:

## Before

screen shot 2016-01-22 at 10 46 10

## After

screen shot 2016-01-22 at 10 46 22


matthew-andrews:
:+1:

PR #864, opened 2016-01-21
Upgrade to Babel v6 for serverside code

callumlocke:
Using similar approach to [next-front-page](https://github.com/Financial-Times/next-front-page/blob/f9e53ba5b767bc609656caa3be9deb6a84416d57/server/app.js) as suggested by @ironsidevsquincy


server/app.js
```diff
@@ -1,2 +1,2 @@
-require('babel/register');
-require('./init.js');
+require('./setup');
```
ironsidevsquincy: good idea, share with tests, I'll do that in the front-page app too

ironsidevsquincy:
:+1:

PR #865, opened 2016-01-22
Special Report streams are looking shoddy

i-like-robots:
- Removes broken background header image.
- Removes broken 'hot topics' take over.

## Before

screen shot 2016-01-22 at 14 06 37

## After

screen shot 2016-01-22 at 14 06 52


server/controllers/v1-taxonomy.js
```diff
@@ -14,7 +14,7 @@ import renderStream from '../lib/render-stream';
 import searchEs from '../lib/search-es';
 import streamConfig from '../config/stream-config';
 import StreamHotTopics from '../lib/stream-hot-topics';
-import streamSpecialReports from '../lib/stream-special-reports';
+// import streamSpecialReports from '../lib/stream-special-reports';
```
andygnewman: Do you want to delete this?

andygnewman:
Just the commented out line, but otherwise :+1:

server/controllers/v1-taxonomy.js
i-like-robots: Happy to. I couldn't find any examples where it worked anymore ¯\(°_o)/¯

PR #869, opened 2016-01-23
Remove role attributes on photo diary images

callumlocke:


andygnewman:
What's you're rationale for the presentational role being incorrect on the photo diary gallery?

callumlocke:
Because they're not presentational, they're content, unless I'm mistaken

andygnewman:
For an accessibility user a photo diary gallery is unlikely to be useful content. @i-like-robots do you have a view?

i-like-robots:
I'm inclined to agree but perhaps the best course of action may be to not assume anything as this is content more than design fluff. I think we should try to mark it up to be as useful as possible.

callumlocke:
> For an accessibility user a photo diary gallery is unlikely to be useful content.

With that logic pretty much every image in the world would have `role="presentation" :)

The `role="presentation"` attribute is for presentational images like spacer gifs or other things that are not content. Images in a 'photo diary' are definitely content.

It's not up to us to hide content from screenreader users, making the executive call that they won't be interested in something about photography because they're probably blind. (1) Blind people still might want to read alt text, follow links and read comments, etc - they can always skip past the section if they're not interested. (2) Not all screenreader users are blind anyway, they might be using it as an assistive device for other reasons.

PR #874, opened 2016-01-26
abstract subheading out of item modelling and include opening

andygnewman:
Improve subheading logic
- use text of first p of opening, if available and no summary
- otherwise, if FastFT and no summary or opening, use text of first p of bodyXML (can deprecate this next week when opening fully ingested)

Will need to redo all this when we do the XSLT on opening.


i-like-robots:
+1

PR #875, opened 2016-01-27
Open Article in Stream - add o-tracking event

andygnewman:
Fires an o-tracking event of page view for a successful article opened in stream. 
Set as an action of inline-view (temporary until happy with how data is being populated into data sources, then revert to view)
Adds to context;
- url set as the pathname of the article being opened
- is_inline boolean flag, set to true
- opened_in set as the pathname of the page in which it was opened in stream


components/card-open-in-stream/inline-article/index.js
```diff
@@ -66,6 +66,17 @@ module.exports = React.createClass({
 		const originalPosition = window.scrollY;
 		this.refs.content.focus();
 		window.scrollTo(0, originalPosition);
+		document.body.dispatchEvent(new CustomEvent('oTracking.event', {
+			detail: {
+				category: 'page',
+				action: 'inline-view',
```
commuterjoy: are we going with `inline-view` now?

andygnewman: Was going to leave it as `inline-view` whilst testing it, but with the intention of swapping back to `view` when we've got it in a state we're happy with.

PR #876, opened 2016-01-27
Turn open in stream off for news feed

andygnewman:
- Change the config
- Differentiate the animation background color depending on whether openable (pink background) or not (off white)


components/stream-item/_main.scss
```diff
@@ -3,7 +3,10 @@
 		background: #f5dfd9;
 	}
 	100% {
-		background: getColor('pink');
+		background: getColor('warm-5');
+		.stream-item--openable {
+			background: getColor('pink');
+		}
```
i-like-robots: can only animate properties, keyframes have no context

matthew-andrews:
:cry:

PR #877, opened 2016-01-27
Concise latest items

sumeetadur:
This provides the option for displaying concise items in the stream. Concise items are displayed without a tag and subheading.


callumlocke:
This looks good to me.

@andygnewman
@i-love-robots

components/card/index.js
```diff
@@ -40,7 +45,7 @@ module.exports = React.createClass({
 						<h3 className="n-card__headline">
 							<a className="n-card__headline__link" href={`/content/${item.id}${item.referrerTracking ? item.referrerTracking : ''}`} data-trackable="headline">{item.title}</a>
 						</h3>
-						<div className="n-card__subheading">{item.subheading}</div>
```
i-like-robots: Does this not simply collapse when there is no content?

sumeetadur: Yes it does, but it leaves a 4px margin-top so thought it was better to remove the element completely so the headline could be exactly vertically-aligned in the middle

sumeetadur:
Closing in favour of https://github.com/Financial-Times/next-stream-page/pull/887

Issue #871, opened 2016-01-25
App architecture


i-like-robots:
We're in a place reminiscent of the articles app around August - we have a build up months of quick hacks, prototypes and tests and now we're approaching a time where we really need to get this stuff specced and more stable.

@callumlocke Did some good work last week towards making our main streams controller more readable  and easier to reason about by lessening the tangle of promises.

@andygnewman Has done some very handy work in how we configure streams of different types which will become increasingly useful.

However we still have a nest of conditionals attaching features. It's not fun to work with!

So, can we discuss how we can better architect this app to make it easier to reason about and much easier to test.

**My idea** is that streams could act a more like streams. We're all familiar with Gulp - input gets passed through filters and transforms then the output is gathered, and I think we could do something similar. 

Features should be standalone modules (mostly the case) and appended to the set of transformations based on the stream config. They'll probably need to implement two methods: the first to modify the request and a second to modify the data sent to the template. Both should return a promise for control flow but whether or not they return a value or modify the given reference is up for debate.

I think this would make for a thin, easily testable controller and each add-on can be reasoned about and tested in isolation. Hardly a conditional in sight.

Anyway, thoughts?

CC master builder @matthew-andrews

matthew-andrews:
I like it.  Would each transform need to block before passing to the next one?

i-like-robots:
_Ideally_ no because each add-on would run in isolation and only be concerned with its business. In reality I'm sure there will be cases where this is not true and concerns are crossed. 100% agree simple and slower > complex and fast.

i-like-robots:
I just pushed a concept of this for fast FT and plugins for tag filters and the article model:

Controller:
https://github.com/Financial-Times/next-stream-page/blob/c4da4699c74ae221aeafd331e9c2c8d3463de169/server/controllers/fast-ft.js

Transform plugin for filters:
https://github.com/Financial-Times/next-stream-page/blob/c4da4699c74ae221aeafd331e9c2c8d3463de169/server/controllers/transforms/filter-by.js

This has been complicated a little by maintaining some backwards compatibility but nothing that wouldn't be easy to clean up.

callumlocke:
I like this a lot.

Could we have a little readme in each transform's directory to say what it does?

callumlocke:
Wondering if a transform should export a single function that *returns* the transformer object with `in` and `out` methods? Some transforms may need to be configured per-stream. 

[...]

PR #879, opened 2016-01-27
New stream controller and stream filters

i-like-robots:
- Prototype implementation of basic stream controller with transforms
- Example article model, basic filter by tag and live stream/open in stream plugins

Implements https://github.com/Financial-Times/next-stream-page/issues/871

Documentation:
https://github.com/Financial-Times/next-stream-page/wiki/Profiles
https://github.com/Financial-Times/next-stream-page/wiki/Transforms


matthew-andrews:
quite neat :+1:

callumlocke:
@i-like-robots how close are you to merging this, do you think? Don't want to rush you, but if it's going to be a while then I need to think about doing our brexit stuff in a different way so I can start showing work-in-progress to stakeholders

i-like-robots:
I'm not going to be working again until Thursday, is that OK? Could a locally running instance + ngrok suffice for a few days? If the filters feature gets wrapped behind a flag then it's safe to go as is.

callumlocke:
OK cool, I'll try ngrok

callumlocke:
re:
> Could a locally running instance + ngrok suffice for a few days? If the filters feature gets wrapped behind a flag then it's safe to go as is.

I can use ngrok for now, but just to make you aware: we have a hard editorial deadline of **next Friday 15th Feb** to get our Brexit stuff online. So the earlier we can get this new stream controller merged into master the better, as ideally I'd like to be able to test our Brexit transforms in production at least a day or two early for testing. (Then we'll just flip the brexitStream flag to on-by-default on Friday.)

callumlocke:
Er I just committed to your branch hope that's OK.

Everything looks good to me now.

i-like-robots:
NP, hopefully the 'next gen' flag won't be around too long =]

PR #880, opened 2016-01-28
add a loading state to open in stream button

andygnewman:
loading


components/card-open-in-stream/_main.scss
```diff
@@ -95,21 +95,22 @@
...
-	cursor: pointer;
-	&:after {
-		@include nextIcon(arrow-down, getColor('grey-tint5'), 10);
-		content: '';
-		position: relative;
-		top: 1px;
-		padding: 0 2px;
-	}
-	.is-open & {
+	&.is-static {
```
i-like-robots: We can just use `&[disabled]` or `&[aria-busy=true]`, they're both 'state' so no need for an extra flag =]

components/card-open-in-stream/inline-article/index.js
```diff
@@ -18,6 +18,7 @@ module.exports = React.createClass({
 		if (!this.state.articleExpanded) {
 			this.setState({
 				articleLoading: true,
+				buttonLabel: 'Loading',
```
i-like-robots: Do we want an ellipsis char (`…`) at the end to create a heightened sense of suspense?

andygnewman: Not a fan, unless they're animated, though happy for a design view. Not sure we have anything similar elsewhere?

components/card-open-in-stream/_main.scss
andygnewman: Good point, will change it.

i-like-robots:
Awesome :ship: (might invert the styles at some point as the button isn't disabled/busy by default)

PR #881, opened 2016-01-28
Incorporate n-barriers in inline article

andygnewman:
- Stop redirect to article page if barrier encountered
- render barrier inline
- import n-barriers styling

Related PRs to

n-barrier: https://github.com/Financial-Times/n-barriers/pull/2

next-product-selector: https://github.com/Financial-Times/next-product-selector/pull/46


commuterjoy:
cool. bit ugly :)

i wonder if we can serve a bespoke, cut down barrier.

matthew-andrews:
can it go full width?

andygnewman:
Just need to have a variant of this for an inline barrier https://github.com/Financial-Times/n-barriers/blob/master/scss/components/barriers/_barrier-grid.scss#L240

that doesn't do L: one-quarter

PR #882, opened 2016-01-29
Fix flexbox layout issues present on 2014 Samsung stock browsers

i-like-robots:
Browser is based on ~mid 2013 version of Blink (shipped in Chrome 28) which has a buggy implementation of flexbox.

The solution is to use less flex container nesting.

Fixes https://jira.ft.com/browse/NFT-305


andygnewman:
:+1:

PR #883, opened 2016-01-29
Fix issue whereby reinstantiating o-date every time a react card comp…

i-like-robots:
…onent was mounted caused IE to break down.

Solution is to not instantiate o-date 50 times at once but let it be instantiated once globally (as it already was) and to init again only when the live stream is updated.

Fixes https://jira.ft.com/browse/NFT-306


matthew-andrews:
:+1:

andygnewman:
D'oh

PR #884, opened 2016-02-01
Show more button for open in stream card should not inherit default b…

i-like-robots:
…ackground colour

Closes https://jira.ft.com/browse/NFT-314


andygnewman:
:+1:

PR #886, opened 2016-02-02
Fix issue whereby a live stream update was closing openable cards.

i-like-robots:
- Uses `shouldComponentUpdate` to only re-render openable cards when necessary (perf++)
- Doesn't re-load article if you've already opened it
- Applies a template property to each stream item so that keys can be generated consistently (fixes https://jira.ft.com/browse/NFT-318)
- Also fixes https://jira.ft.com/browse/NFT-316 by switching from `window.scrollY` to `window. pageYOffset`


components/card-open-in-stream/inline-article/index.js
```diff
@@ -15,46 +15,54 @@ module.exports = React.createClass({
 	},
 
 	handleClick () {
-		if (!this.state.articleExpanded) {
+		if (this.state.articleLoading) {
+			return;
```
andygnewman: button is disabled at this point so should be impossible for this to happen

components/card-open-in-stream/inline-article/index.js
```diff
@@ -15,46 +15,54 @@ module.exports = React.createClass({
...
+				this.fetchArticle();
+			}
 		}
 	},
 
 	articleClosing () {
 		this.setState({
 			articleExpanded: false,
-			buttonLabel: 'Show more',
-			inlineArticle: null
```
andygnewman: Presume this is to retain the article, so don't need to fetch it again if reopened

components/card-open-in-stream/inline-article/index.js
```diff
@@ -64,10 +72,10 @@ module.exports = React.createClass({
 			buttonLabel: 'Show less',
 			inlineArticle: html
 		});
-		const originalPosition = window.scrollY;
```
andygnewman: This and the scrollTo on line 69 were to overcome the scroll action that the focus would undertake.

When you reinstate it you could make the window.scrollY IE compatible document.documentElement.scrollTop.

components/card-open-in-stream/inline-article/index.js
i-like-robots: I was just being belt and braces to cover all states. I think it is actually possible to click twice (state is set async) but this wouldn't save us from that anyway.

components/card-open-in-stream/inline-article/index.js
i-like-robots: Yep!

components/card-open-in-stream/inline-article/index.js
i-like-robots: window.pageYoffset FTW

andygnewman:
:+1:

PR #889, opened 2016-02-03
Concise items (as a transform)

sumeetadur:
Provides the option for displaying concise items in the stream. Concise items are displayed without a tag and subheading.


server/transforms/concise-items.js
```diff
@@ -0,0 +1,14 @@
+module.exports = (removeTags = true) => {
+	return {
+		post(req, res) {
+			res.templateData.items = res.templateData.items.map(item => {
+				item.isConcise = true;
+				item.subheading = null;
+				if (removeTags === true || item.primaryTag && removeTags.indexOf(item.primaryTag.name) > -1) {
```
matthew-andrews: please use IDs rather than the tag names for comparisons ^^

callumlocke:
Closing this PR but keeping the branch alive as we're going to re-base it onto something else.

PR #892, opened 2016-02-04
Decouple transform series-ing from controller/profiles

callumlocke:
@i-like-robots 

following on from slack message, here is what I'm suggesting.

Differences:

- each profile now has a just single `transform` property, not an array.
- there's a new lib function `reduceTransforms(t1, t2...)` which collapses a list of transforms into one.
  - the reduced transform runs all the individual transforms in series, including setting up the private scope thing.
- it should also be a little bit faster, as assembling the series now only happens once, at startup, instead of on every request.

What problem this solves:
- the logic for wrapping a series of transforms (and setting up a private scope for each) is now decoupled from the profiles and from the controller, and can therefore be reused easily in tests.
  - (the existing `filter-by.test.js` doesn't need this, because the filter-by transform happens not to use `this`, so it can be used directly without wrapping... but other transforms will need to have their `this` set up)

Not sure about the name `reduceTransforms` though, it doesn't adequately convey that it will set up a shared `this` for your input transform's pre and post methods... Maybe something like `prepareTransforms` would be better.


callumlocke:
this has a design flaw: although the scope is private per-transform, it is shared across all requests

to fix this, the promise series does need to be assembled on every request after all; it's not possible to just do it once at startup :disappointed:

callumlocke:
OK this is now a simpler change than originally described.

All this really does now is simplify the transfrom-wrapping logic and move it into its own file so it can be reused by tests.

i-like-robots:
I like the change to move the promise series out of the controller. Re. solving the scoping - especially when under test - maybe we could just specify transforms as a class or constructor function and instantiate new instances for each request? Let the language do the work =]

callumlocke:
down with classes

callumlocke:
I mean, I'm down with classes, not "down with classes"

matthew-andrews:
i'm ok-ish with classes but try to avoid inheritance :grin: :see_no_evil:

i-like-robots:
I've done some investigation with classes and... this is better IMO, certainly when it comes to authoring transforms.

```js
// transforms/live-stream.js
const register = require('./register');

class LiveStream {
	pre(req) {
		this.isFirstPage = req.perPage.page === 1;
	}

	post(res, flags) {
		res.templateData.isLive = this.isFirstPage && flags.liveStreams;
		res.templateData.isAlive = res.templateData.isLive ? true : res.templateData.isAlive;
	}
}

module.exports = register(LiveStream);

// register.js
module.exports = (Transform) => {
	return (config) => () => new Transform(config);
};

// config/stream-plugins.js
const profiles = require('./profiles');
const series = require('promise-patterns').series;

module.exports = (req, res) => {
	const profile = profiles.find(profile => profile.match(req, res));
	const transforms = profile.transforms.map(transform => transform());

	return {
		pre: () => series(transforms.map(
			transform => Promise.resolve(transform.pre && transform.pre(req, res.locals.flags))
		)),
		post: () => series(transforms.map(
			transform => Promise.resolve(transform.post && transform.post(res, res.locals.flags))
		))
	};
};
```

callumlocke:
interesting. I just pushed at the same time as your comment came in, taking a slightly different approach:

- there's `new TransformSeries(transforms)` to flatten a bunch of transforms into one.
- there's no 'register' step, but the transform module exports a function that returns a class like this:

```js
module.exports = function () {
  return class {

  }
}
```

callumlocke:
updated my comment (hit post too early) - the config gets passed in when you 'get' the class.

i.e.

```js
  transforms: [
    injectThing(someConfig) // returns a *configured* InjectThing class
  ]
```

i-like-robots:
But the transform class is not being provided with it

https://github.com/Financial-Times/next-stream-page/blob/callum/resolve-transforms/server/lib/transform-series.js#L9

callumlocke:
It doesn't need it. If your transform class needs configuring, you just put your class definition inside your function, so it can see the config:

```js
// server/transforms/inject-thing.js

module.exports = function (configGoesHere) {
  return class InjectThing {
    pre() {...}
  }
}
```

i-like-robots:
Ah, yes that will do =]

i-like-robots:
:+1:

callumlocke:
also this way means you never need to bother writing a constructor function that has to manually attach args to 'this' on every instantiation

i-like-robots:
Less misdirection and magic is always a good thing :sparkler:

callumlocke:
are you happy with this PR then? can you merge it into your branch?

I'm currently working on Brexit stuff on a branch off `matth/fast-ft-filters`, so if you merge this PR into yours then I can rebase my feature branch off yours and carry on with this classes approach...

i-like-robots:
Go for it

PR #895, opened 2016-02-08
[WIP] Brexit stuff

callumlocke:
This PR adds two transforms: 'stickies' and 'meta-card'.

This is not currently very presentable (I've left ugly outlines to help with layout debugging as it's still a WIP) but it's all behind the `brexitStream` flag, which is off-by-default, so should be safe to merge.

**stickies**:

- This transform adds a 'stickies' component into the stream. The component is a single stream-item, but contains multiple content items (like photo-diary).
- Deduping - regular stream items get removed if they are duplicates of anything inside a 'stickies' component.
- The contents are loaded from a remote list, either from Bertha or the CAPI. There's a lib function that abstracts this away.
- A 'stickies' component is laid out in a 'triplet' or 'quartet' or 'list' layout, depending on whether it's got 3, 4 or more items respectively.
- If you set `critical: true`, then in the event of failure to load the list of sticky items, it will 500 the whole page, preferring to serve stale from the CDN than to serve a page without the stickies.

**meta-card**:

- This transform injects a fragment for the custom IG insert for the Brexit stream.
- The fragment comes from a URL controlled by IG, and gets sanitized just in case.

**additional changes in this PR**:

- cleared out junk from early brexit experimentation
- changed the `make watch` task to make it watch `./components` as well as `./client`


components/meta-card/_main.scss
```diff
@@ -0,0 +1,13 @@
+.meta-card {
+	background-color: white;
```
matthew-andrews: either add a TODO to pull from o-colors or pull from o-colors plz

components/stickies/sticky-item/_main.scss
```diff
@@ -0,0 +1,51 @@
...
+
+	&:not(:last-of-type) { border-bottom: 1px solid #f6e9d8; }
+}
+
+.sticky-item__image {
+	display: none;
+}
+
+.sticky-item__link {
+	color: #333333;
```
matthew-andrews: palette?

components/stickies/sticky-item/_main.scss
```diff
@@ -0,0 +1,51 @@
+.sticky-item {
+	padding: 10px;
+	outline: 1px solid teal; // TODO debugging
```
matthew-andrews: color

package.json
```diff
@@ -28,6 +29,7 @@
     "ft-next-express": "^15.0.0",
     "ft-next-logger": "^4.0.0",
     "intl": "^1.0.0",
+    "n-image": "^2.2.2",
```
matthew-andrews: probably ought to be be bower rather than package

views/article-list.html
```diff
@@ -69,6 +55,12 @@ <h2 class="stream__title">{{longTitle}}</h2>
 	</div>
 </div>
 
+{{#if isBrexit}}
+{{!-- TEMPORARY - any misc hacks for brexit can go here, until they're can be done properly --}}
```
matthew-andrews: since we don't need this… can we remove?

matthew-andrews:
:+1:

package.json
callumlocke: Tried this, proxyquire got confused. Not sure if there's any way to make proxyquire aware of bower_components as a place to look for modules. I've reverted to putting n-image in package.json.

components/image-container/_main.scss
```diff
@@ -0,0 +1,23 @@
...
+	overflow: hidden; // in case image turns out not to have precisely the expected ratio (eg due to rounding error)
+
+	> .n-image {
+		position: absolute;
+		left: 0;
+		top: 0;
+		width: 100%;
+		min-width: 100%;
+		max-width: 100%;
+		display: block;
```
i-like-robots: no block context when positioned absolutely

components/meta-card/index.js
```diff
@@ -0,0 +1,36 @@
+import React from 'react';
+
+export default class MetaCard extends React.Component {
+	
+	constructor() {
+		super();
+		this.state = { expanded: false };
```
i-like-robots: Should be `getInitialState` maybe?

components/stickies/sticky-quartet/index.js
```diff
@@ -0,0 +1,22 @@
+import React from 'react';
+import StickyItem from '../sticky-item';
+
+export default class StickyQuartet extends React.Component {
+	render() {
```
i-like-robots: components with only a render method can be simple functions, just in case you got tired of typing classes =]

package.json
i-like-robots: We got around this in other places with a couple of tricks:

1. only putting the require call within a block that will run on the client, E.G. `componentDidMount`
2. https://github.com/Financial-Times/next-stream-page/blob/38a5405263d688ac812a03b0debe1e4bfe579698/test/components/stream.test.js#L9

server/config/profiles.js
```diff
@@ -27,6 +29,44 @@ module.exports = [
 		],
 	},
 
+	{
+		// UK's EU Referendum (Brexit)
+		match: req => req.params.thingId === 'NDdiMzAyNzctMTRlMy00Zjk1LWEyZjYtYmYwZWIwYWU2NzAy-VG9waWNz',
+		transforms: [
+			// TODO: this is just a hack to get 'isAlive'
+			liveStream(),
```
i-like-robots: This will start a poller, I guess that's not needed. What is 'alive' being used for, just instantiating some client side stuff?

components/meta-card/index.js
sumeetadur: I'm doing it this way as we're going with ES6, extending React.Component instead of using React.createClass

i-like-robots: You're right, not seen [this caveat](https://facebook.github.io/react/blog/2015/01/27/react-v0.13.0-beta-1.html) before!

server/config/profiles.js
callumlocke: The only thing we're using alive-ness for is expanding/unexpanding the MetaCard. Which is just toggling a class on that component, based on clicking the 'show more/less' button inside it. Should we do it another way, and make it not alive?

components/meta-card/index.js
```diff
@@ -0,0 +1,36 @@
...
+
+		if (this.state.expanded) {
+			classes += ' meta-card--expanded';
+			buttonText = 'Close referendum guide';
+		} else {
+			buttonText = 'For full referendum guide';
+		}
+
+		return (
+			<div className={classes}>
```
i-like-robots: `aria-expanded`

components/meta-card/index.js
```diff
@@ -0,0 +1,36 @@
...
+				<div className="meta-card__main-content" dangerouslySetInnerHTML={{__html: item.mainHTML}}></div>
+				<div className="meta-card__extra-content" dangerouslySetInnerHTML={{__html: item.extraHTML}}></div>
+				<div className="button" onClick={this.toggle}>{buttonText}</div>
```
i-like-robots: Should be a button with `aria-controls` attr

components/meta-card/index.js
callumlocke: ah. nice. i'll do that

server/config/profiles.js
i-like-robots: The main overhead of using isAlive is that all the state must be injected into the page and reinitialised. I guess that's quite a bit of extra overhead of a toggle button. That said, dropping the code into an arbitrary client-side only file seems   weird and it will get lost. We're using o-expander quite extensively, maybe hook into that?

components/meta-card/index.js
i-like-robots: There's one here with loading state too https://github.com/Financial-Times/next-stream-page/blob/master/components/card-open-in-stream/inline-article/index.js#L140-L149

server/config/profiles.js
callumlocke: :+1: o-expander, i'll set that up

callumlocke:
On Matt's advice (can't remember which), I got rid of all clientside React for this.

Decided instead to use o-expander data attributes inside the fragment itself.

This means the 'MetaCard' component is now really just a wrapper for a single fragment of HTML, so I renamed it `CustomFragment`.

matthew-andrews:
it was hinchliffe i think, but sounds sensible

callumlocke:
I did a rebuild without cache, but it still failed with something weird and bowery, so I'm seeing if this fixes it...

matthew-andrews:
yup fair enough.  good to go now?

PR #899, opened 2016-02-09
Update related tags integration for compatibility with V3 index

i-like-robots:
https://github.com/Financial-Times/next-ft-api-client/pull/59 fixed this feature which had been failing since we deleted the V1 index. Well, here it is again...

screen shot 2016-02-09 at 17 32 03


server/controllers/v1-taxonomy.js
```diff
@@ -217,7 +217,7 @@ export default coroutine(function *(req, res, next) {
 
 		// Decorate with related tags
 		if (stream.taxonomy && stream.tag.term.name) {
-			const tags = yield relatedTags(stream.taxonomy, stream.tag.term.name, req.params.thingId);
+			const tags = yield relatedTags(stream.tag.term.name);
```
andygnewman: Is it valid to use name in this circumstance rather than id?

i-like-robots: You're right, the label may be ambiguous. I'll need to update the client module first.

i-like-robots: Sorted here: https://github.com/Financial-Times/next-ft-api-client/pull/65

i-like-robots:
Published next-ft-api-client v3.21.1, just waiting for that to propagate.

matthew-andrews:
really nicely done :+1:

PR #902, opened 2016-02-11
Rehydration re-arrangement

i-like-robots:
This PR does two things:

1. Moves dehydrated data into a script element so that it doesn't have to be escaped. This slims the page down by around ~10%.
2. Moves the `templateData.items` property into a `stream` namespace. This is so other data (I.E. filters and facets) can be more easily passed to client-side React stuff.


andygnewman:
:+1:  LGTM

callumlocke:
:+1:

PR #904, opened 2016-02-11
Brexit continued

callumlocke:


callumlocke:
Would @andygnewman or @i-like-robots be able to take a look at this?

components/card/index.js
```diff
@@ -45,7 +50,7 @@ module.exports = React.createClass({
 								<ODate dateTime={item.publishedDate} />
 							</div>
 						</div>
-						<div className="n-card__meta">{actions}</div>
+						{actions}
```
callumlocke: removed the extra wrapper as it seems to be a mistake – double-nesting `n-card__meta` divs causes an o-grid-column nested inside another o-grid-column, doubling up gutters. 

also the 'official' template doesn't double-nest them: https://github.com/Financial-Times/next-card/blob/master/templates/card.html

components/stickies/sticky-quintet/_main.scss
```diff
@@ -0,0 +1,101 @@
...
+		width: (3/7) * 100%;
+		margin-right: -20px;
+	}
+
+	.sticky-quintet .sticky-item--4 .sticky-item__image {
+		display: block;
+		margin-bottom: 15px;
+	}
+
+	.sticky-item--5 .sticky-item__subheading { display: none; }
```
i-like-robots: visually hide so it stays accessible?

components/stickies/sticky-quintet/_main.scss
```diff
@@ -0,0 +1,101 @@
...
+		&:after {
+			content: "";
+			display: table;
+			clear: both;
+		}
+	}
+
+	.sticky-quintet__others-pair-1 {
+		float: left;
+		width: (4/7) * 100%;
```
i-like-robots: `percentage(4 / 7)`

components/stickies/sticky-quintet/_main.scss
callumlocke: do we have a preferred next mixin/technique for visual hiding?

i-like-robots: No mixin but this should be the main technique we use https://github.com/Financial-Times/n-util/blob/master/main.scss#L41-L50

bower.json
```diff
@@ -16,8 +16,7 @@
     "n-message-prompts": "Financial-Times/n-message-prompts#^v2.0.0",
     "n-notification": "^2.0.0",
     "n-topic": "^5.0.0",
-    "n-barriers": "^1.0.0",
-    "o-colors": "^3.3.1"
```
matthew-andrews: intentional?

components/stickies/index.js
```diff
@@ -1,26 +1,27 @@
...
+
+		if (layoutHint === 'featured') {
+			if (items.length > 5) items.length = 5;
+
+			inner = <StickyQuintet items={items} />
+
+			// TODO: switch down to Quartet or Triplet or Singlet layout if it's fewer than 5 items
+			// (but for now, the Quintet layout can handle it by just leaving empty spaces)
 		}
+		else inner = <StickyList items={items} />
```
matthew-andrews: lol i didn't know you could do this.  javascript is nuts.

test/server/transforms/branding.test.js
```diff
@@ -40,8 +41,8 @@ describe('Branding transform', () => {
 
 		xit('sets various branding properties?', () => {})
```
matthew-andrews: `xit`?

bower.json
matthew-andrews: ah think you're using some speical next colour getting mixin instead,

matthew-andrews:
:+1:

test/server/transforms/branding.test.js
callumlocke: no idea, that's in master though, not in this PR

matthew-andrews: good point…  carry on

components/stickies/index.js
i-like-robots: The linter should get angry at this for a lack of semi-colon. I guess when it sees JSX it just doesn't bother checking the line.

PR #906, opened 2016-02-12
Fix rss and json endpoints

matthew-andrews:
Not sure how to test :sweat_smile:


i-like-robots:
Shouldn't be too hard with the `node-mocks-http` modules as used in the [controller spec](https://github.com/Financial-Times/next-stream-page/blob/master/test/server/controllers/stream.test.js)

(I also fudged this into a [WIP branch](https://github.com/Financial-Times/next-stream-page/commit/d34c5492865baa9aa4a7a73be4bcc4b2d722abb3) but we should fix this ASAP!)

PR #907, opened 2016-02-12
Client side filters

i-like-robots:
fastft

Integrates basic filters into our Reactified streams. Works by passing the primitive `fetchItems` and `updateItems` methods of the `Stream` component into the `StreamFilters` component which simply updates the URL via `pushState` before calling them.


andygnewman:
:+1:

PR #908, opened 2016-02-12
Switch the news feed stream to use the new streams controller

i-like-robots:
Adds support to new stream controller for requests without a TME tag. Also upgrades Nock because, wow.


matthew-andrews:
:+1:

PR #910, opened 2016-02-15
Put custom fragment back behind flag

callumlocke:


callumlocke:
@matthew-andrews I inadvertently removed the flag check for this feature

matthew-andrews:
ok :+1:

PR #913, opened 2016-02-17
Brexit additional featured items

sumeetadur:
adds more sticky items halfway down the page on Brexit stream:

image


components/stickies/sticky-list/_main.scss
```diff
@@ -1,4 +1,70 @@
-// TODO
 .sticky-list {
-	display: none;
+	display: flex;
+	flex-wrap: wrap;
+	padding-top: 20px;
+	.sticky-item {
+		border-bottom: none;
```
callumlocke: change this to `border-bottom: 0` to fix lint error

components/stickies/sticky-list/_main.scss
```diff
@@ -1,4 +1,70 @@
...
+		padding-left: 0 !important;
+	}
+	button.o-expander__toggle {
```
callumlocke: and change `button.o-expander__toggle` to just `.o-expander__toggle` to fix the lint error

sumeetadur: Thanks, already fixed these but I've been making more changes before committing..a good reminder to have smaller commits..

callumlocke:
:+1:

@andygnewman @matthew-andrews can you take a look?

components/stickies/sticky-list/index.js
```diff
@@ -1,13 +1,42 @@
...
-				{items.map((item, i) => <StickyItem item={item} key={i} itemNumber={i} />)}
+			<div className="sticky-list o-expander" data-o-component="o-expander" data-o-expander-shrink-to="hidden" data-o-expander-expanded-toggle-text="Show fewer featured articles" data-o-expander-collapsed-toggle-text="Show more featured articles">
+
+				<div className="sticky-list__chunk">{chunks.shift()}</div>
+
+				<div className="o-expander__content" aria-hidden="true">
+					{chunks.map(chunk => <div className="sticky-list__chunk">{chunk}</div>)}
+				</div>
+
+				<button className="o-expander__toggle o--if-js"></button>
```
andygnewman: Don't forget your data-trackable on the button

andygnewman:
General comment IMHO - not sure why you would want an expander. Expanders should be used for less important optional content, whereas featured content feels like it should always be shown. What the right amount to feature is an open question though.

callumlocke:
I tend to agree, but I think the editorial intention for these stickies is not exactly to 'feature' important stories, but just to have something to break the flow with a few evergreen items with nice pictures. The first four are always visible, and then you expand to get some more.

PR #914, opened 2016-02-17
Show branding in stickies

callumlocke:
looks like this:

image


lc512k:
:+1:

components/stickies/sticky-item/_main.scss
```diff
@@ -21,6 +21,16 @@
 	margin: 7px 0;
 }
 
+.sticky-item__branding {
+	> a {
```
andygnewman: the nLinksTopic mixin is our standard way of doing tag links.

callumlocke: yeah I was going to use that, but apparently we've been given the ok from next design team to try out this design in the brexit stickies:

image

(haven't actually done the big quote marks yet, but that will come)

callumlocke:
thanks!

components/stickies/sticky-item/_main.scss
andygnewman: We already use this type of styling in the article app (eg. https://next.ft.com/content/a97efd1e-d34c-11e5-829b-8564e7528e54) and it's on the home page (usually found in the opinion section).

PR #917, opened 2016-02-18
Live stream notice

i-like-robots:
online:

screen shot 2016-02-18 at 11 02 55

offline:

screen shot 2016-02-18 at 11 36 33

So our users don't have to manually refresh the page every 6 seconds =]

(Also removes the long-ignored stream sorting component now that the flag has expired and been deleted.)


andygnewman:
:+1:

PR #918, opened 2016-02-18
Remove dedicated article styling 

andygnewman:
Post incorporation of n-content-body, this strips out the dedicated styling that was previously doing the job of styling the open in stream article.


i-like-robots:
:+1:

PR #928, opened 2016-02-26
Fix deduping

callumlocke:
Deduping sticky articles isn't working, meaning this page has some repeated headlines:

https://next.ft.com/stream/topicsId/NDdiMzAyNzctMTRlMy00Zjk1LWEyZjYtYmYwZWIwYWU2NzAy-VG9waWNz


commuterjoy:
:+1: looks ok, with the caveat that it's me saying that.

callumlocke:
thanks, you look trustworthy enough

i-like-robots:
<nag>test?</nag>

callumlocke:
Yes you are quite right
And that would have avoided this bug in the first place
Maybe I will write one

callumlocke:
@i-like-robots I've added a test for the stickies transform. Can you have a look?

i-like-robots:
:ship:

PR #934, opened 2016-03-01
Use methode lists

callumlocke:
This is a small change to make the Brexit stream load its 'sticky' articles from a Methode list, i.e. via the CAPI instead of Bertha.

The CAPI list-loading logic was already written and [tested](https://github.com/Financial-Times/next-stream-page/blob/58868f77bbcebcee9fdc20819a8042b4e0bcbc9b/test/server/lib/load-list.test.js), hence why this PR doesn't have many changes in it.


i-like-robots:
:ship:

PR #936, opened 2016-03-02
Brands branding using brands API

andygnewman:
Objectives:

Style header for branded pages (with entries in the brands API developed by UPP).

To be behind a flag `streamBrandsDesign`.

Still To Do:
- [x] Work with Editorial to identify initial brand pages to go live with and tidy up data
- [x] Figure out how to get the custom background image into the production repo
- [x] Route only those brands selected with Editorial through to the new controller

Example Styling

Mobile
screen shot 2016-03-02 at 14 39 48

Desktop
screen shot 2016-03-02 at 14 38 52


server/transforms/branding.js
```diff
@@ -1,10 +1,37 @@
...
+
 module.exports = brand => {
 	return class {
 		post(res) {
-			// TODO: go and talk to the branding API!
-			res.templateData.branding = brand;
+			if (brand) {
+				res.templateData.branding = {};
+				// leaves customised branding fastFT and brexit as they are
+				return res.templateData.branding.class = brand;
```
i-like-robots: Let's make this one line into its own transform with a big deprecated warning

server/transforms/branding.js
```diff
@@ -1,10 +1,37 @@
...
+			} else {
+				const id = res.templateData.conceptIdV2;
+				if (id) {
+					const path = `/brands/${encodeURIComponent(id)}`;
+					return capiV2({ path: path })
```
i-like-robots: `({ path })`

server/transforms/branding.js
```diff
@@ -1,10 +1,37 @@
...
+						res.templateData.branding = response;
+						let imageUrl = res.templateData.branding._imageUrl;
+						if (imageUrl) {
+							imageUrl = encodeURIComponent(imageUrl);
+							res.templateData.branding._imageUrl = `https://image.webservices.ft.com/v1/images/raw/${imageUrl}?source=next&width=150&height=150&quality=highest`;
+						}
+						res.templateData.branding.class = 'branded';
+					})
+					.catch(
+						error => logger.warn('Fetching branding API failed.', error.toString())
```
i-like-robots: Only return when we want to return, may cause lint warnings in future 

```
error => {
  logger.warn()
}
```

server/transforms/concordance.js
```diff
@@ -0,0 +1,45 @@
...
+const capiV2 = require('next-ft-api-client/src/utils/capi-v2');
+
+module.exports = (direction) => {
+	return class {
+		pre(req) {
+			const id = req.params.thingId;
+			if (direction === 'TMEtoV2') {
+				req.params.thingIdTME = id;
+				const path = `/concordances?authority=http://api.ft.com/system/FT-TME&identifierValue=${id}`;
+				return capiV2({ path: path })
```
i-like-robots: `({ path })`

views/partials/sub-header.html
```diff
@@ -1,15 +1,27 @@
-<div role="navigation" aria-label="category navigation" class="sub-header sub-header--{{branding}}" data-trackable="sub-header">
+<div role="navigation" aria-label="category navigation" class="sub-header{{#if branding.class}} sub-header--{{branding.class}}{{/if}}" data-trackable="sub-header">
```
i-like-robots: might blow up if `branding` is `undefined`

client/components/sub-header/_main.scss
```diff
@@ -132,10 +132,41 @@
...
+	@include oGridRespondTo($until: M) {
+		.sub-header__page-title {
+			display: block;
+			text-align: center;
+			margin-right: 0;
+		}
+
+		.n-myft-ui--follow {
+			display: table;
+			margin: auto;
```
i-like-robots: `0 auto`

client/components/sub-header/_main.scss
```diff
@@ -155,6 +186,42 @@
...
+.sub-header__image-wrapper {
+	text-align: center;
+	@include oGridRespondTo(M) {
+		position: absolute;
+		right: 30px;
+		bottom: -35px;
+	}
+}
+
+.sub-header__image--main {
```
i-like-robots: I think `--main` variation is not required here

client/components/sub-header/_main.scss
```diff
@@ -155,6 +186,42 @@
...
+	margin-bottom: 10px;
+}
+
+.sub-header__description {
+	margin-top: 10px;
+	@include oGridRespondTo(M) {
+		margin-top: 0;
+	}
+	p {
+		margin-bottom: 0;
```
i-like-robots: By default there will be a top margin too, should this be set/removed too?

client/components/sub-header/_main.scss
```diff
@@ -164,7 +231,8 @@
 	}
 
 	&[data-o-grid-colspan] {
-		margin-bottom: 20px;
+		// margin-bottom: 20px;
+		position: initial;
```
i-like-robots: Does this do anything? `initial` in IE

server/transforms/concordance.js
```diff
@@ -0,0 +1,45 @@
...
+
+// HACK from v1-v2-taxonomy
+const capiV2 = require('next-ft-api-client/src/utils/capi-v2');
+
+module.exports = (direction) => {
+	return class {
+		pre(req) {
+			const id = req.params.thingId;
+			if (direction === 'TMEtoV2') {
+				req.params.thingIdTME = id;
```
i-like-robots: Just realised this is not quite right... could this be a `post` transform and apply the results straight to the template data?

i-like-robots: > not quite right 

As in that's not the request that came in... so we could generate incorrect links etc.

PR #941, opened 2016-03-03
No need to hit 'things' endpoint to fetch a tag

i-like-robots:
Currently on the site we hit CAPI to fetch the tag data:

https://github.com/Financial-Times/next-stream-page/blob/05e366be09120d2b4deac7d36910d9b7ac0c1d20/server/controllers/stream.js#L10-L24

I changed this to use API client:

https://github.com/Financial-Times/next-stream-page/commit/9c02622979b3a56635881375a0e26c860d1e70b8

But the API client changed again:

https://github.com/Financial-Times/next-ft-api-client/pull/68

And it now just hits ES, which is redundant since we'll be fetching content from there anyway. (This also broke the specs in my branch because of missing mocks).


andygnewman:
LGTM :+1:

PR #943, opened 2016-03-03
Revert "Stop rogue-including fonts"

matthew-andrews:
Reverts Financial-Times/next-stream-page#940

reopening this so that the tests pass whilst we deicde what to do


callumlocke:
closing as this was fixed by #944 instead

PR #946, opened 2016-03-04
Kill /pages/ URLs — WIP

matthew-andrews:


i-like-robots:
Ooh, yay! I never understood these anyway... :ship:

matthew-andrews:
there's a couple of editorial requirements we gotta kill first :(

PR #947, opened 2016-03-04
Move non-shared components back into client directory.

i-like-robots:
I think we got a bit too excited. Also shifts Sass partials around to be alongside JS components. Makes it a little easier to navigate around the project and understand how it's evolving.


callumlocke:
photo-diary component – I agree with putting its Sass and JS together in one place, but shouldn't it be in shared? it's a react component that gets server-rendered, isn't it?

everything else :+1:

i-like-robots:
Photo diary is still in shared components, just moved from stream/ -> photo-diary/ =]

PR #949, opened 2016-03-04
Apply: Border use case for fastFT color for page menu styling

andygout:
cc @i-like-robots @andygnewman 

Uses more semantically correct `border` use case, added to `next-sass-setup` in this PR: https://github.com/Financial-Times/next-sass-setup/pull/116


i-like-robots:
:+1:

PR #950, opened 2016-03-08
Sticky list item CSS

sumeetadur:
Fixes the padding and border styles for the Brexit sticky list on all screen sizes

screen shot 2016-03-08 at 11 27 23
screen shot 2016-03-08 at 11 27 06


sumeetadur:
@i-like-robots / @matthew-andrews could you have a look please?

PR #951, opened 2016-03-08
Brexit - Fix padding in sticky quintet

sumeetadur:


sumeetadur:
@i-like-robots small change, ready to ship

PR #953, opened 2016-03-10
Stickies configured from Bertha

### LINKED ISSUE #932: Using stickies on other streams
The newsroom wants to reuse the Brexit 'stickies' feature on the page for the [UK Budget stream](https://next.ft.com/stream/topicsId/YmRlZTE5ZTAtNjMzMS00OGIxLTk5ZWEtMGI5YmY4NzViYmVi-VG9waWNz) in 2 weeks. ...any thoughts?

If we start adding stickies to other streams, I'm thinking the transformation config (currently in [profiles.js](https://github.com/Financial-Times/next-stream-page/blob/0ab95c980095821dbe49133926fc0ac5024bebab/server/config/profiles.js)) should probably come from a polled Bertha sheet, rather than requiring a code deployment for any change... something like:

listid | streamid | index | style
----|-----|-----|-----
7f12a510-bf62-11e5-9fdb-87b8d15baec2 | NDdiMzAyNzctMTRlMy00Zjk1LWEyZjYtYmYwZWIwYWU2NzAy-VG9waWNz | 0 | quintet
305def14-b3a1-11e5-8358-9a82b43f6b2f | NDdiMzAyNzctMTRlMy00Zjk1LWEyZjYtYmYwZWIwYWU2NzAy-VG9waWNz | 7 | list

or am I thinking about this the wrong way?

cc @matthew-andrews @i-like-robots @andygnewman @ironsidevsquincy

callumlocke:
Fixes #932.

This makes the stickies transform run for all streams. The transform itself then decides whether to actually do anything by checking the current stream ID against a polled manifest.

Here is the manifest:
https://docs.google.com/spreadsheets/d/1-ieW0eSgpQYP_iu6_8oibHb3TuUUKiMoZnQ1yJ0LHNE/edit#gid=0


test/server/transforms/lib/request-response.js
```diff
@@ -2,8 +2,8 @@
 
 const httpMocks = require('node-mocks-http');
 
-module.exports = (params, data, page) => {
-	let req = httpMocks.createRequest(params || {});
+module.exports = (options, data, page) => {
+	let req = httpMocks.createRequest(options || {});
```
callumlocke: see https://github.com/howardabrams/node-mocks-http#createrequest for why this name change

callumlocke:
Also, as this change opens up the possibility of adding stickies to any stream (and we're planning to use it for the US 2016 Election stream very soon), having this feature behind the `brexitStream` flag no longer makes sense. Shall I make a new flag called `streamStickies?`

i-like-robots:
New flag sounds good. Happy to have this listed as 'default' as there's no overhead really. Might be a little superfluous to run on blogs (inc. FastFT) but I'm sure lots of people will want it.

callumlocke:
@matthew-andrews are you ok with this change too? Including new streamStickies flag?

matthew-andrews:
i'm cool if matt/h's
 cool

Issue #922, opened 2016-02-22
State containers?


i-like-robots:
We've built a bunch of React components and we've thrown together features quickly. It's time soon for a refactor so we can better apply all the things we've learned... and maybe that means learning some more.

With the top-level streams component also implementing live streams and soon to include some complex filtering components there's a lot going on. Currently this is pieced together by passing callbacks around in an ad-hoc manner which is OK but terrifying to anybody unfamiliar with the code and will almost certainly create problems as we try to extend the code further.

Ideally we should be able to control the stream items and filters more easily and I reckon that means implementing some state containers. I think it will make our components easier to reason about, more resilient to refactors and easier to test.

Looking around it's clear Redux is the leanest and most popular implementation at this time.

andygnewman:
Sounds like a plan - is this the Redux you're referring to? https://github.com/reactjs/redux

i-like-robots:
That's the one. I think it would be ideal if we _didn't_ need such a tool but I haven't worked out anything better. I'm not actually bothered which flavour of state container we go for but Redux seems well documented and has React-specific optimisations, although it does need an add-on for async operations.

callumlocke:
I love redux. Only downside is the ecosystem is a bit fragmented from the outset, but that's a reflection of how popular it is.

redux-thunk... one reservation I have is I don't feel like the redux thunk/promise middleware really solves the async problem. OK it allows you to pass thunks or promises around, but I feel like this just patches over the synchronous nature of the reducers system, and still requires you to worry about what happens when things happen in different orders. It looks like [redux-saga](https://github.com/yelouafi/redux-saga) might be a good answer to this, asking you formally separate your async logic into an async system that lives side-by-side with your redux store/reducer logic, but I haven't tried it yet...

callumlocke:
after a little more reading on redux-saga, I've started to feel it may be a solution in search of a problem. It's not too hard to think of (most) async actions as two separate actions that each cause a synchronous state mutation, like "start loading thing" and "display thing".

i-like-robots:
Sounds reasonable. We have to handle the inbetween loading state anyway (displaying the right indications, setting `aria-busy=true` or whatever).

i-like-robots:
I'm currently working on implementing redux for the root stream component and child live stream and stream filter components. So far so good.

PR #954, opened 2016-03-10
Stream redux

i-like-robots:
- Wraps stream into a Redux provider
- Moves data for stream and stream filters component into shared Redux store
- Extracts out live stream functionality into separate component
- Extracts autocomplete functionality into separate component
- Removes the clear all filters button from stream filters component as requested
- When adding/removing a filter it fetches the stream before appending/removing the tag from UI/URL
- ... which also means you can't queue requests = fewer errors

## Why do all this?

- No more tangle of callbacks and props
- Standardised, shared busy and invalid states

## Caution

This is the first time I've tried to use Redux. It probably isn't quite right and retro-fitting it was a bit of a pain.

CC https://github.com/Financial-Times/next-stream-page/issues/922


shared/reducers/stream.js
```diff
@@ -0,0 +1,91 @@
...
+function messages(state, action) {
+	state = state || [];
+
+	switch(action.type) {
+		case 'ADD_MESSAGE':
+			return state.concat(action.payload);
+		default:
+			return state;
+	}
+}
```
callumlocke: could use ES6 default args for all these reducers

```js
function messages(state = [], action) {
  switch(action.type) {
```

callumlocke:
haven't tried running it, but this looks great

PR #959, opened 2016-03-15
improved tracking markup

wheresrhys:
and preparing for vanity urls @i-like-robots


server/controllers/pages.js
```diff
@@ -27,6 +27,7 @@ module.exports = function(req, res, next) {
 				path: req.path,
 				title: results.title,
 				longTitle: longTitle(results.title),
+				trackablePageName: 'curated',
```
matthew-andrews: bike shedding but not sure this is quite right.  not all pages are curated and not all normal streams are un-curated.

matthew-andrews:
the controllers that are responsible for most streams (`stream.js`, `v1-taxonomy`, and `v1-v2-taxonomy.js`) are missing from this PR, is that intentional?

server/controllers/pages.js
wheresrhys: we already distinguish between curated and stream pages in tracking based on url. Even if the terms aren't 100% reliable as a description, it's potentially still useful as pages aren't feature-equivalent with streams. 

Arguably I should use 'page', but I think that'd get confusing. Could go with 'preditor'? Either way, the work to expose this info to o-tracking is worth doing for search vs stream anyway

wheresrhys:
yes (as they already set the taxonomy and concept ids needed in the template), but I haven't really thought much about passing page type to spoor. Will mull it over and then they will change

server/controllers/pages.js
matthew-andrews: `legacy-page` 

these are deprecated now and will be killed soonish so it doesn't matter too much.  :smile:

wheresrhys:
gonna get page type from appName (now sent with every event), with a tiny bit of rewriting in spoor-consumer

PR #960, opened 2016-03-15
Use React component for section-meta in custom fragment

sumeetadur:


sumeetadur:
@callumlocke

sumeetadur:
@callumlocke could you have a look? small change but just want to make sure

callumlocke:
This looks good to me. Have pinged @andygnewman in case he has an idea about the very long import line you were concerned about.

shared/components/custom-fragment/index.js
```diff
@@ -1,11 +1,12 @@
 import React from 'react';
+import SectionMeta from '../../../bower_components/n-section/components/section/section-meta/section-meta';
```
callumlocke: @i-like-robots Andy said (on Slack) it's unusual to pull JS out of Bower components like this. Our intention here is to avoid duplicating the DOM structure for this element, but if it's not generally done then maybe there's a better approach. Can you advise?

i-like-robots: I'd duplicate for as this approach is likely fragile. How we deal with these primitive building blocks going forward is yet to be decided...

I imagine it will involve bundling them into a monolith-ish UI repo so there is visibility and searchability over the components and their relationships.

sumeetadur: Thanks, I'll close this PR now

PR #963, opened 2016-03-17
Update section-meta in custom fragment to match next-front-page

sumeetadur:


sumeetadur:
@i-like-robots small change & only affecting Brexit stream, so will merge this...any objections?

PR #966, opened 2016-03-21
Initialise video component for inline articles

adgad:
/cc @andygnewman @i-like-robots 

Initialises the video component for articles opened in stream.


shared/components/card-open-in-stream/inline-article/index.js
```diff
@@ -15,6 +15,15 @@ module.exports = React.createClass({
 		};
 	},
 
+	componentDidUpdate() {
```
adgad: Not sure if this is the best place in the lifecycle for it. Initially had it in articleLoaded - but on second open it seemed to lose the race condition.

test/shared/components/inline-article.test.js
```diff
@@ -7,14 +7,21 @@ require('./utils/dom');
...
 // so finally let's get React and its test utils.
 const React = require('react');
 const TestUtils = require('react-addons-test-utils');
 
 let instance;
 
-const Subject = require('../../../shared/components/card-open-in-stream/inline-article');
+const videoInitStub = sinon.stub();
+const Subject = proxyquire('../../../shared/components/card-open-in-stream/inline-article', {
+	'n-video': { init: videoInitStub }
```
i-like-robots: stub is a React component so `{ init: () => {} }` is probably better or `{ init: sinon.stub() }`

test/shared/components/inline-article.test.js
```diff
@@ -7,14 +7,21 @@ require('./utils/dom');
 // use Chai for expectations
 const expect = require('chai').expect;
 // and Sinon for spies and stubs (with a wrapper for use with React classes)
+const sinon = require('sinon');
 const ReactSinon = require('./utils/react-sinon');
+// only render the component under test by stubbing its dependencies
+const stub = require('./utils/stub-component');
```
i-like-robots: unused!

andygnewman:
I guess it may be useful in future to have a flag on the article in the ingest that says whether there is a video included, so we only init nVideo when there is a video. Or indeed for anything where we end up initing client side JS.

PR #970, opened 2016-03-23
Better custom fragments

callumlocke:
In the same vein as #953 (Stickies configured from Bertha), this PR makes custom fragments configurable from Bertha and puts them behind their own dedicated flag, `customStreamFragments`.

The manifest: https://docs.google.com/spreadsheets/d/13N3rLTt_mcGJQA2aXEXgeVd0J0JIeqoarR2CGnIdnio/edit

This is more complicated than the Stickies code, it involves two steps of polling, so I've made a `CustomFragmentsPoller` class to abstract this away and to make it more testable. But when running the app there is only ever one instance of this class. It polls the manifest (linked above) and maintains a collection of sub-pollers for individual fragment endpoints.

The transform is now synchronous and adds virtually no overhead to requests, so I hope it's OK to add this one as a default transform.

Please review – @matthew-andrews @andygnewman @i-like-robots


andygnewman:
I'm not sure I agree with adding it to the default transforms as it adds a dependency into all other streams regardless of whether they need it or not.

Whilst I understand the point about overhead to requests (which we need to be very careful about given the app's current memory performance), my other concern is that it adds a further dependency to all stream pages. If a problem developed within that code base it has the potential to bring down all streams whilst only bringing benefit to a couple of streams currently.

It may be that when we really get into the streams work that using this as a default may be a useful way to achieve our objectives, but there are other ways such as the work we're doing on blogs branding where the data is being fed via bertha into the API and we're using an api call to get the configuration data / content.

callumlocke:
> my other concern is that it adds a further dependency to all stream pages

I totally agree, which is why I've been very careful to write this in a way that adds virtually **zero** overhead to any stream that doesn't use it (and in fact, very very little overhead to streams that do use it). It's literally just an object look-up – if the current stream ID is not a key in that object, nothing happens.

> If a problem developed within that code base it has the potential to bring down all streams whilst only bringing benefit to a couple of streams currently.

I don't think that's possible. The transform just looks up the stream's ID in an object. If that object doesn't have that stream ID as a key, the result will be undefined and nothing happens.

While I'm fairly certain the getData() call could never error under any circumstances, I'd be happy to make this change to alleviate any doubt:

```diff
 import CustomFragmentsPoller from '../lib/custom-fragments-poller';
 
 const poller = new CustomFragmentsPoller(
   'https://bertha.ig.ft.com/view/publish/ gss/13N3rLTt_mcGJQA2aXEXgeVd0J0JIeqoarR2CGnIdnio/manifest'
 );
 
 poller.start();
 
 let count = 0;
 
 export default () => {
   return class {
     pre(req, flags) {
       if (flags.customStreamFragments && req.perPage.offset === 0) {
-        this.fragments = poller.getData(req.params.thingId);
+        try {
+           this.fragments = poller.getData(req.params.thingId);
+         }
+         catch (error) {
+           // use ft logger to log it and carry on
+         }
       }
     }
 
     post(res) {
       if (!this.fragments) return;
 
       for (const { index, html, bleed, title } of this.fragments) {
         res.templateData.stream.items.splice(index, 0, {
           bleed, title,
           fragment: html,
           template: 'custom-fragment',
           id: count++
         });
       }
     }
   };
 };
```

callumlocke:
Just re-read your comment and realised my first paragraph is irrelevant as you weren't talking about the overhead :) but the rest is relevant

server/transforms/custom-fragments.js
```diff
@@ -0,0 +1,38 @@
...
+					this.fragments = poller.getData(req.params.thingId);
+				}
+				catch (error) {
+					logger.error({ event: 'CUSTOM_FRAGMENTS_GET_DATA_ERROR', error });
+				}
+			}
+		}
+
+		post(res) {
+			if (!this.fragments) return;
```
andygnewman: Is this a more effective way round than doing

```
if (this.fragments) {
    do fragments stuff
}
```

server/transforms/custom-fragments.js
```diff
@@ -0,0 +1,38 @@
...
+poller.start();
+
+let count = 0;
+
+export default () => {
+	return class {
+		pre(req, flags) {
+			if (flags.customStreamFragments && req.perPage.offset === 0) {
+				try {
+					this.fragments = poller.getData(req.params.thingId);
```
andygnewman: I think you're right that we don't want to log every time this fails as it will fail most of the time if put onto every stream. Though this will make any debugging harder.

server/transforms/custom-fragments.js
callumlocke: imo this style makes the intention clearer (return without modifying anything) and looks cleaner (less indenting and bracing), but I guess it's subjective...

matthew-andrews: i think both are valid/clear approaches.  pr reviews best used to focus on functionality rather than style :innocent:.  perhaps go with callum's for now?

callumlocke:
By the way, I've been talking with Luke this morning about eventually removing all this stuff entirely, making it redundant, configuring streams by putting things in Elastic Search instead, which would be much more scalable. I think there's an issue by matt h about this somewhere. But that's obviously a bigger conversation, and in the meantime we need a reasonable way to get custom items into other streams (need to do it for the US Election stream asap).

server/transforms/custom-fragments.js
andygnewman: I guess I didn't phrase my comment well. It wasn't so much a style point more a question over truthiness vs not operator for which I understand there can be some subtleties about. Was more a learning point for me than critique of the code.

callumlocke: Oh I see what you're saying. I think the truthiness vs actual trueness factor is the same either way here... The ! operator guarantees a boolean, but then so does just putting something in an if condition... if(truthy) is identical to if(true). So that's not a reason for me doing it this way round here. It was purely because I thought it looked nice :)

matthew-andrews:
we should have a little think about whether elasticsearch vs. a no-sql db vs. something else.

i-like-robots:
:+1:

PR #971, opened 2016-03-24
Add referrer to fastft inline views

wheresrhys:
@andygnewman @commuterjoy Is this preferable to the status quo (where referrer for each inline article would be the page somebody got to fastft from, which seems like a massive distortion of the facts)


andygnewman:
:+1:

PR #972, opened 2016-03-24
Only show sticky-item image div when it has an image

sumeetadur:
It's cleaner and also fixes this bug:

screen shot 2016-03-24 at 16 52 50

After fix:

small screen:
screen shot 2016-03-24 at 17 05 54
medium screen:
screen shot 2016-03-24 at 17 06 07
large screen:
screen shot 2016-03-24 at 17 06 14


sumeetadur:
@callumlocke

i-like-robots:
LGTM

Aside: Have you had any thoughts about reconciling this component with the homepage team as they have something very similar?

sumeetadur:
Thanks, will look into that

PR #973, opened 2016-03-29
Stream facets layout 2.0

i-like-robots:
- Autoresize text input
- Click anywhere in box to focus text input
- Suggestions follow text input
- Uses tag facets API for much better suggestions!

### Little screen

screen shot 2016-03-29 at 11 17 28

### Big screen

screen shot 2016-03-29 at 11 22 45


andygnewman:
LGTM

Issue #930, opened 2016-03-01
Determine output format by query string rather than suffix


matthew-andrews:
It'll help us simplify how we do vanity URLs.

https://next.ft.com/stream/brandId/NTlhNzEyMzMtZjBjZi00Y2U1LTg0ODUtZWVjNmEyYmU1NzQ2-QnJhbmRz.rss
=>
https://next.ft.com/stream/brandId/NTlhNzEyMzMtZjBjZi00Y2U1LTg0ODUtZWVjNmEyYmU1NzQ2-QnJhbmRz?format=rss

https://next.ft.com/stream/brandId/NTlhNzEyMzMtZjBjZi00Y2U1LTg0ODUtZWVjNmEyYmU1NzQ2-QnJhbmRz.json
=>
https://next.ft.com/stream/brandId/NTlhNzEyMzMtZjBjZi00Y2U1LTg0ODUtZWVjNmEyYmU1NzQ2-QnJhbmRz?format=json

The logic we have that looks at the `accept` header should be unaffected by this change.

i-like-robots:
+1

matthew-andrews:
i think this is fixed now? @i-like-robots

i-like-robots:
Yep. We still support the dot extensions. I assume we need to remove those and add a redirect?

matthew-andrews:
i think we could probably redirect them through in `next-redirector`

`/(.*).(json|rss)` => `/$1?format=$2`

i-like-robots:
yep yep

PR #974, opened 2016-03-29
Fix paths

i-like-robots:
1. RSS link was not specified for new stream pages.
2. We cannot rely on the pathname when pages have a vanity URL. Currently live streams and filters are broken as they require a non-existent `.json` vanity also. This should be considered a temporary work around while an approach is decided for dealing with this long term.


server/controllers/stream.js
```diff
@@ -53,6 +53,7 @@ module.exports = (req, res, next) => {
 
 		res.templateData = {
 			stream: { items, total: items.total, name: tag.prefLabel },
+			path: `/stream/${tag.taxonomy}Id/${tag.idV1}`,
```
wheresrhys: something bit more descriptive maybe? Not sure what. `internalUrl`, `streamPath`?

i-like-robots: This is just for compatibility as it's broken. Will add TODO note.

i-like-robots:
CC https://github.com/Financial-Times/next-stream-page/issues/930

PR #977, opened 2016-03-30
remove slug library

callumlocke:


andygnewman:
I think you'll need to change image-container/index.js also as you'll have the full url, so you're not using n-image.

https://github.com/Financial-Times/next-stream-page/blob/master/shared/components/image-container/index.js#L25

something like;
```
let src = `"${props.url}?source=next&amp;fit=scale-down&amp;width=110&amp;compression=best&amp;format=png&amp;quality=highest&amp;dpr=2";
return (
    <img
        className={className}
        alt={props.alt}
	src={src} />
);
```

andygnewman:
Or as you've just done keep it self contained if the behaviour of image-container needs to be retained for other use cases.

callumlocke:
should it be `source=next`? i did `source=next-stream-page`

callumlocke:
I've changed the source to `next` and squashed. Think this is ready, if anyone can review.

andygnewman:
I've always just used `next` rather than the full app name, though I guess either is legitimate.

sumeetadur:
LGTM :+1:

andygnewman:
:+1:  (assuming it passes the tests)

PR #978, opened 2016-03-30
Fix support for .rss extension and add test

i-like-robots:


matthew-andrews:
i'm really sorry about this, i totally forgot

PR #979, opened 2016-03-31
Small autocomplete fixes

i-like-robots:
As reported in https://jira.ft.com/browse/NFT-398

- Bumps tag autocomplete version (https://github.com/i-like-robots/react-tags/pull/6)
- Clears stale suggestions when selected tags are added/removed


andygnewman:
:+1:

PR #983, opened 2016-04-01
New index is case sensitive so use camel-cased specialReports taxonomy

i-like-robots:


wheresrhys:
:+1:

matthew-andrews:
let's get this shipped first but can we add a smoke test for this? e.g. ensure this reutrn a non 0 number of items ? special-reports.json

wheresrhys:
^^ or some health checks which query ES for this and other crucial tags

PR #985, opened 2016-04-01
New pagination

i-like-robots:
The pagination currently sits alone in the long ignored stream 'footer'. This brings it into the stream and won't break when a user filters the page.

screen shot 2016-04-01 at 16 46 57


server/lib/format-query-string.js
```diff
@@ -0,0 +1,8 @@
+'use strict';
+
+module.exports = (params) => {
+	const chunks = Object.keys(params)
+		.map(item => `${item}=${encodeURI(params[item])}`);
+
+	return `?${chunks.join('&')}`;
```
ironsidevsquincy: could just add the `join` to the above and directly return

shared/components/pagination/index.js
```diff
@@ -0,0 +1,27 @@
+'use strict';
+
+const React = require('react');
+
+module.exports = (props) => {
```
ironsidevsquincy: is this not actually a React component then?

i-like-robots: Yep, it's just a 'stateless' one.

ironsidevsquincy:
:+1:

andygnewman:
LGTM 👍

PR #991, opened 2016-04-04
Special reports transforms

i-like-robots:
Let the new streams controller handle the listing of individual special reports.

Also switches podcast transforms to only run on the first page.


test/server/transforms/special-report.test.js
```diff
@@ -0,0 +1,54 @@
...
+	beforeEach(() => {
+		subject = new Subject();
+	});
+
+	describe('transforms the response', () => {
+
+		context('for a special report stream', () => {
+
+			beforeEach(() => {
+				res = requestResponse(null, { taxonomy: 'specialReports', conceptId: 'abc123' }).pop();
```
andygnewman: I'm liking the use of pop() here.

test/server/transforms/special-report.test.js
```diff
@@ -0,0 +1,54 @@
...
+				expect(res.templateData.isSpecialReport).to.be.true;
+			});
+
+			it('corrects generated path', () => {
+				expect(res.templateData.path).to.equal('/special-reports/abc123');
+			});
+
+		});
+
+		context('not for a special report stream', () => {
```
andygnewman: the context and description are a little confusing.

Could be

`context('for a stream that is not a special report', `
`it('does not flag the stream as a special report',`

andygnewman:
Minor comment on test descriptions, but otherwise 👍

test/server/transforms/special-report.test.js
i-like-robots: updated

PR #992, opened 2016-04-05
Add markets data transform for organisation streams

i-like-robots:
This pulls the existing markets data module into a transform for organisation streams. This won't currently do anything as they're not being served on the new architecture but it will when we migrate.

Before migration we will also need to clean up the separate organisations header partial because we're mothballing the organisation info for the foreseeable. Currently there are no tests for the underlying markets data module so I've been quite descriptive in the test for this transform with the intention of refactoring later.

Running on a V2 stream:

screen shot 2016-04-05 at 12 12 10


andygnewman:
Great 👍

PR #993, opened 2016-04-05
Ftlabs tldr (show first 5 sentences in stream)

JakeChampion:
For an oft-raised use case:
I want to get the news as quickly and efficiently as possible, with minimal effort

Hypothesis:
By pressuring readers to click through to full article pages, we may be harming the experience of a core of hard-core readers. Can we demonstrate (and measure) the value of encouraging these readers *not* to click through to articles?


server/models/article.js
```diff
@@ -16,9 +17,18 @@ module.exports = opts => {
 			return item;
 		}
 
+		let summary;
+		if (wantsSummary) {
+			const amountOfSentences = 5;
```
andygnewman: You could use cheerio to do this work of getting the text out more accurately than a regexp.

Eg. https://github.com/Financial-Times/next-stream-page/blob/90dc3d6b1f8104045f90e4fec7bd690a8c8634f3/server/models/subheading.js#L10-L12

server/models/article.js
```diff
@@ -16,9 +17,18 @@ module.exports = opts => {
...
+			const articleText = item.bodyXML.replace(/<(?:.|\n)*?>/gm, ' ');
+			const articleSentences = articleText.split('.');
+			summary = articleSentences.slice(0, amountOfSentences).join('.\n') + '.';
+		}
+
 		return {
 			id: item.id,
 			title: item.title,
+			summary: summary,
```
andygnewman: can just do

```
title: item.title,
summary,
publishedDate: item.publishedDate,
```
where the name of the assignment is the same as the assignee.

shared/components/card/index.js
```diff
@@ -29,7 +29,9 @@ module.exports = React.createClass({
 				</div>;
 		}
 
-		if (item.subheading) {
+		if (item.summary) {
```
andygnewman: could do something more like;

let subheadingContent = item.summary || item.subheading;
subheading = <div className="n-card__subheading">{subheadingContent}</div>;

server/lib/summary.js
```diff
@@ -0,0 +1,12 @@
+'use strict';
+
+const cheerio = require('cheerio');
+
+module.exports = item => {
+	const amountOfSentences = 5;
+	const articleText = cheerio.load(item.bodyXML)('p').text();
+	const articleSentences = articleText.split('.');
```
andygnewman: You should be able to do this all in one
```
return cheerio.load(item.bodyXML)('p').map.(function(i, el) {
    if (i < amountOfSentences) {
        return $(this).text();    
    }
}).get().join('\n');
```

andygnewman:
Jake - have you also seen the openingXML which is being added to some of the articles through CAPI?

It's an editorially curated TLDR type concept (only applies to Wordpress articles like fastFT at the moment).

We're going to expose this on the fastFT stream as part of the open in stream, just waiting till we get it all transformed to HTML - next week.

for the TLDR it may be best to use the opening if we have it and then only do the TLDR as you have it if there isn't one.

server/lib/summary.js
JakeChampion: This looks like it would return a combination of paragraphs. I'm looking to return a set of sentences.

PR #994, opened 2016-04-05
Define ad slots inside the app

adgad:
/cc @i-like-robots @andygnewman @VladDubrovskis @andrewgeorgiou1981 
PR to define ad slots and styles within the app and remove them from the über next-ads-component.


views/article-list.html
```diff
@@ -11,6 +11,7 @@
 <link rel="alternate" type="application/rss+xml" href="{{path}}?format=rss" />
 {{/defineBlock}}
 
+
```
VladDubrovskis: is this the only change to this file? :)

VladDubrovskis:
:+1:

PR #996, opened 2016-04-06
n.makefile

matthew-andrews:
As a separate PR I will switch to using `n-makefile` for builds… which is way faster!


andygnewman:
LGTM

package.json
```diff
@@ -47,9 +47,13 @@
     "signed-aws-es-fetch": "^1.0.0"
   },
   "devDependencies": {
+    "bower": "^1.7.9",
```
i-like-robots: Could we move this to `optionalDependencies`?

i-like-robots: So local install can use the `--no-optional` flag

matthew-andrews: good idea

PR #999, opened 2016-04-06
Remove old v1 and v2 controllers and related bits

i-like-robots:
- Removes V1 controller, sends all TME streams to new controller
- Removes V1/V2 controller and redirects requests for V2 UUID streams to equivalent TME tag
- Cleans up no longer used modules
- Removes organisation info from sub header


server/init.js
```diff
@@ -51,8 +32,8 @@ app.get('/special-reports/:thingId.?:extension(rss|json)?', require('./controlle
 app.get('/special-reports/:specialReportId/special-reports.?:extension(rss|json)?', require('./controllers/related/special-reports'));
 app.get('/special-reports/:specialReportId/:relatedTaxonomy.?:extension(rss|json)?', require('./controllers/related/tag'));
 
-app.get('/stream/:thingType(primarySectionId|genreId|brandId|peopleId|sectionsId|organisationsId|regionsId|topicsId|authorsId|subjectsId|icbId|iptcId)/:thingId.?:extension(rss|json)?', require('./controllers/v1-taxonomy'));
-app.get('/:thingType(people|organisations)/:thingId.?:extension(rss|json)?', require('./controllers/v1-v2-taxonomy'));
+// V2 UUID to V1 TME
+app.get('/:thingType(people|organisations)/:thingId.?:extension(rss|json)?', require('./controllers/v2-redirect'));
```
andygnewman: Presumably somewhere in next we're choosing to use the V2 id for these thingTypes in links etc. Would it be worthwhile changing this at source?

andygnewman:
Just one small comment, though wouldn't expect that to be fixed in this PR. 👍

server/init.js
i-like-robots: Yes where it's not too inconvenient to do so...

PR #1002, opened 2016-04-06
Ftlabs tldr for people/organisations/pages streams

JakeChampion:


andygnewman:
I wouldn't worry about pages streams - there's a plan to deprecate them shortly and as you've seen the people & organisation streams have already migrated to the new controller.

PR #1003, opened 2016-04-07
bring summary into overall model

andygnewman:
@JakeChampion - just a couple of changes just to bring summary in line with how stream is structured.

There should also be a test for the summary module.


JakeChampion:
LGTM :+1: I'll look at tests for this now

PR #1005, opened 2016-04-07
Code coverage

i-like-robots:
Implements code coverage using [nyc](https://github.com/bcoe/nyc) (a sort of front for Istanbul that can handle our Babel setup).


bjfletcher:
\o/

When running `make coverage`, I see the test results but not coverage results. Is it that they go into `coverage/coverage.json`? It'd be nice to see a formatted output for the coverage results.

i-like-robots:
A nice coverage report can be generated by switching the `--reporter` option to `lcov` but this implementation generates a JSON to be sent to Coveralls when running the master branch. I can try and tweak this so it's possible to run the `coverage` task locally without making any changes to the Makefile.

Issue #1000, opened 2016-04-06
Post refactor refactors


i-like-robots:
Now we are running on the new architecture we can do some more clean up. Some of the original modules may lack tests so this is a good opportunity to spec these things.

- [x] Integrate get current nav module into the [transform](https://github.com/Financial-Times/next-stream-page/blob/90dc3d6b1f8104045f90e4fec7bd690a8c8634f3/server/transforms/current-nav.js)
- [x] Integrate markets data module into the [transform](https://github.com/Financial-Times/next-stream-page/blob/f6cd697b5644d344c899aecf4f233b92fbd38cee/server/transforms/markets-data.js)
- [x] Integrate hot opinion module into the [transform](https://github.com/Financial-Times/next-stream-page/blob/master/server/transforms/hot-opinion.js) 
- [x] Integrate hot analysis module into the [transform](https://github.com/Financial-Times/next-stream-page/blob/master/server/transforms/hot-analysis.js)

PR #1006, opened 2016-04-07
Integrate modules into transforms and new tests

bjfletcher:
See https://github.com/Financial-Times/next-stream-page/issues/1000 for all the glory details.


test/server/transforms/current-nav.test.js
```diff
@@ -4,29 +4,107 @@
...
+	// 	subject.post(res);
+	//
+	// 	expect(res.templateData.currentNav).to.deep.equal({
+	// 		parent: {
+	// 			"name": "Energy",
+	// 			"id": "MzA=-U2VjdGlvbnM=",
+	// 			"href": "/stream/sectionsId/MzA=-U2VjdGlvbnM="
+	// 		}
+	// 	});
+	// });
```
bjfletcher: @i-like-robots See the line over there -> https://github.com/Financial-Times/next-stream-page/pull/1006/files#diff-a6a5ae09b99ba92a0a7d6ada3fdfeb53L10 and the TODO above that. That means the test here (above) fails. What should we do about this? If the logic is right, then is it ok to remove the test above? Do you envision us needing the above test again in the future?

i-like-robots: I don't think this functionality is required... we already know the 'last level' because that is the page currently being rendered, right?

server/transforms/hot-stuff.js
```diff
@@ -1,8 +1,16 @@
...
+import hotOpinionT from './hot-opinion';
 import StreamHotTopics from '../lib/stream-hot-topics';
-import streamHotOpinion from '../lib/stream-hot-opinion';
-import streamHotAnalysis from '../lib/stream-hot-analysis';
+import hotAnalysisT from './hot-analysis';
+
+// TODO (Ben): Look at this more...
+// 1. should we do series of transforms in profile instead?
+// 2. is HotOpinion thread safe?
+const HotOpinion = hotOpinionT();
```
bjfletcher: @i-like-robots See my questions above.

server/transforms/hot-stuff.js
```diff
@@ -1,8 +1,16 @@
...
+const hotOpinion = new HotOpinion();
+const HotAnalysis = hotAnalysisT();
+const hotAnalysis = new HotAnalysis();
```
bjfletcher: Previously, with `require`, we could do it nicely as follows:

```
const HotOpinion = require('./hot-option');
const hotOpinion = new HotOpinion();
```

but with `import`, I'm having to do all that in 3 lines instead of 2 lines, as above. It's not very readable. Any suggestion?

i-like-robots: I prefer to use `require` for now... I don't mind which you prefer

server/transforms/hot-stuff.js
i-like-robots: 1. Hot analysis and hot opinion will _always_ be together
2. What is it that's concerning you?

test/server/transforms/current-nav.test.js
bjfletcher: @i-like-robots Not sure I follow. Do you mean we should remove:

`if (!(currentNav && currentNav.children)) currentNav = false;`

or should we leave it?

i-like-robots: @andygnewman ?

server/transforms/hot-stuff.js
bjfletcher: @i-like-robots I see that `pre` sets some state that `post` then picks up. Potential scenario:

```
const t = new HotOpinion();
t.pre(req1);
t.pre(req2);
t.post(res1); // fail - it uses state resulting from req2 not req1
```

we'd need to create a whole new transformation for each `req` and `res` pair to avoid the above scenario.  But that's a bit expensive and we'd need to ensure that the API enforces that.  

Wouldn't it be better to return state from `pre` and then pass it in to `post`? This pure FP/immutable way of doing things I find is safer.

i-like-robots: > we'd need to create a whole new transformation for each req and res

We are currently creating new instances of the required transforms for each request. In an earlier iteration we did not use a class, the transform modules exported the `pre` and `post` methods and they were called with a shared scope (possibly `res.locals`) which is a little more efficient but had potential for unexpected collisions.

i-like-robots: The implementation for this is here: https://github.com/Financial-Times/next-stream-page/blob/master/server/lib/transform-series.js

bjfletcher: Thanks :) Yes, I saw that. I've adopted that pattern for now. See https://github.com/Financial-Times/next-stream-page/pull/1006/commits/50e6c9eb455bd390424b4eeff5071a73ce9bc420

A bit uneasy about leaving state inside transforms. Maybe we can discuss about this more later.

i-like-robots: Sure thing, any improvements welcome!

server/transforms/hot-stuff.js
```diff
@@ -1,28 +1,39 @@
...
+const HotAnalysis = require('./hot-analysis')();
+const StreamHotTopics = require('../lib/stream-hot-topics');
 
 module.exports = () => {
+
 	return class {
+
+		constructor() {
+			this.hotOpinion = new HotOpinion();
+			this.hotAnalysis = new HotAnalysis();
```
i-like-robots: Maybe these transforms should be integrated into this one? The two will always be mutual.

test/server/transforms/current-nav.test.js
```diff
@@ -4,29 +4,97 @@
...
+						},
+						{
+							"name": "Utilities",
+							"id": "MzM=-U2VjdGlvbnM=",
+							"href": "/stream/sectionsId/MzM=-U2VjdGlvbnM="
+						}
+					]
+				}
+			]
+		});
```
i-like-robots: Is there any added benefit in performing the deep equals over simply checking the expected properties exist? E.G. 

```js
expect(res.templateData.currentNav).to.include.keys('name', 'id', 'href', 'children');
```

i-like-robots:
Very thorough tests =]

bjfletcher:
@i-like-robots Thanks :) I've just done screenshot comparison of UK and World pages between `master` an this branch, no regression. So I think this is good to go (*famous last words* no doubt...)

server/transforms/current-nav.js
```diff
@@ -1,15 +1,44 @@
...
+						name: parent.name,
+						id: parent.id,
+						href: parent.href
+					};
+				}
+				return true;
+			} else if (item.children) {
+				return find(item.children, id, item);
+			}
+			return false;
```
i-like-robots: No need to return a value from this function as the result is never assigned

server/transforms/hot-stuff.js
i-like-robots: Yeah, definitely pull the hot opinion and hot analysis transforms into this... they're of little or no use on there own

server/transforms/markets-data.js
```diff
@@ -1,8 +1,63 @@
...
+				if (response.data.items && response.data.items.length) {
+					return formatMarketsData(response.data.items.pop());
+				}
+			})
+			.catch(error => {
+				if (!(error instanceof fetchres.BadServerResponseError)) {
+					raven.captureMessage(error);
+				}
+			});
+};
```
i-like-robots: no need for trailing `;`

server/transforms/markets-data.js
```diff
@@ -1,8 +1,63 @@
...
+			.then(function(response) {
+				if (!response.ok) {
+					logger.warn('Failed getting markets data for', symbol);
+				}
+
+				return response;
+			})
+			.then(fetchres.json)
+			.then(response => {
+				if (response.data.items && response.data.items.length) {
```
i-like-robots: doe we need to check for the `response.data` property first in case the data is malformed?

server/transforms/hot-stuff.js
```diff
@@ -19,10 +46,12 @@ module.exports = () => {
 				const items = res.templateData.stream.items;
 				const conceptId = res.templateData.conceptId;
 				res.templateData.hotTopics = new StreamHotTopics(items, conceptId).process();
-				res.templateData.hotOpinion = streamHotOpinion(items);
-				res.templateData.hotAnalysis = streamHotAnalysis(items);
-				res.templateData.stream.items = res.templateData.stream.items.splice(0, this.count);
+				res.templateData.hotOpinion = findHotOpinion(items);
+				res.templateData.hotAnalysis = findHotAnalysis(items);
+				res.templateData.stream.items = res.templateData.stream.items.slice(0, this.count);
```
i-like-robots: I just noticed that `this.count` is `undefined`, so I guess this is really `res.templateData.stream.items.slice(0)`

bjfletcher: @i-like-robots Yes, `undefined` has the same behaviour as 0. I'll make it explicit.

bjfletcher: @i-like-robots BTW, what did you do to see that it was `undefined`?

i-like-robots: Nothing, I just noticed it when checking the diff

i-like-robots:
squash and merge! 🚢

PR #1007, opened 2016-04-08
Featured Content Transform + Component

sumeetadur:
Makes the "Sticky List" a transform + component separate from the stickies. See https://github.com/Financial-Times/next-stream-page/pull/913

screen shot 2016-04-08 at 17 09 04


sumeetadur:
@callumlocke

andygnewman:
In your PR can you explain the proposition that you are trying to achieve.

sumeetadur:
@andygnewman @matthew-andrews can you have a look?

andygnewman:
I'm confused on the naming in this PR.

I'm used to an interstitial being a [transitionary page] (https://en.wikipedia.org/wiki/Interstitial_webpage), but would I be right in thinking that it in this context it is just a block of content? Could you name as something more descriptive, eg. curatedContent.

Similar with the article component. It isn't actually the full article is it? Elsewhere across Next where we show an outline of an article we refer to it as a 'card'. Could you name this highlightedCard or similar?

The code itself looks OK.

callumlocke:
Thanks Andy. You're right the naming isn't right yet, we've just been chatting about how best to rename things...

- We're thinking 'featured content' may be the least bad name for this component. So the flag would be `streamFeaturedContent`, and the component named `<FeaturedContent>`, and the transform `featuredContent`. (Technically it might not always be curated; the source of articles may be some other dynamically generated list.)

- Yes HighlightedCard is better, we'll rename it. Might be worth merging it into Card in future so there's just one n-card-powered way of linking to articles.

andygnewman:
Am liking the new suggested names.

andygnewman:
I believe @andygout has been doing some work to 'origamify' the card styles used on the home page.  Probably not worth doing much on that until we get into the full stream redesign work over the next few weeks.

andygout:
Yup - `n-card` is going to essentially serve as a wrapper for `o-card`. As @andygnewman says, am currently making `n-card` (and also `n-section` for the rankings we use for Most Read on frontpage) compatible with `o-card`.

sumeetadur:
@andygout, thanks I'll make the transition to n-card when it's ready.

@andygnewman could you take a look at the recent changes I've made to see if it's all good to go? Thanks

server/transforms/featured-content.js
```diff
@@ -0,0 +1,89 @@
+import api from 'next-ft-api-client';
+import cheerio from 'cheerio';
+import extractImage from '../lib/extract-image';
+import getBranding from 'ft-n-article-branding';
+import legacyTagCompat from '../models/legacy-tag-compat';
+import loadList from '../lib/load-list';
+import Poller from 'ft-poller';
+import subheading from '../models/subheading';
+import { logger } from '@financial-times/n-logger';
```
andygnewman: See https://github.com/Financial-Times/next-stream-page/pull/1029

matthew-andrews: 😄

server/transforms/featured-content.js
```diff
@@ -0,0 +1,89 @@
...
+					);
+
+					res.templateData.stream.items.splice(data.index, 0, data);
+				}
+			});
+		}
+	}
+}
+
+function featuredContentItemModel(item) {
```
andygnewman: There's a lot of duplication between this function and the [article model] (https://github.com/Financial-Times/next-stream-page/blob/master/server/models/article.js) used in streams. Just a thought as to whether we could merge them? If you don't do it now, could you raise an issue against the stream app for this work to be done in future.

Additionally we have an [issue](https://github.com/Financial-Times/next-es-interface/issues/354) raised against next-es-interface, that we should add flags to content that describe what it contains. Can you add this as a use case.

sumeetadur: Thanks @andygnewman, I'll add an issue for merging the models and reply to the other issue. Let me know if this PR is good to go.

andygnewman:
LGTM 👍

PR #1013, opened 2016-04-11
update the targeting of top and middle ads based on the new propositi…

VladDubrovskis:
…on naming and put it behing the flag

/cc @adgad @andrewgeorgiou1981


server/controllers/stream.js
```diff
@@ -72,7 +84,8 @@ module.exports = (req, res) => {
 			title: tag.prefLabel,
 			taxonomy: tag.taxonomy,
 			conceptId: tag.idV1,
-			layout: 'article-list'
+			layout: 'article-list',
+			adsLayout: getAdsLayout(adsLayout, flags)
```
adgad: Do we actually need any of this adsLayout? It doesn't look like we're using it, and just using the flag instead.

adgad: We do use this, for the n-layout parameter (set in n-layout).

adgad:
:+1:

PR #1014, opened 2016-04-11
Improve live filters

i-like-robots:
Just a few quick things to improve filters:

- Don't clear suggestions on each input change unless blank
- Deliver more suggestions to utilise inline filtering
- Fix Redux warnings re. unexpected property


andygnewman:
👍

PR #1016, opened 2016-04-12
pick up stream id from the data attribute rather than window.location

VladDubrovskis:
/cc @adgad @i-like-robots


i-like-robots:
👍 

Just to re-cap the discussion: in an ideal world we would have re-written this into a server-side component already but as we're re-developing the streams app it's likely it will be written out soon.

PR #1017, opened 2016-04-12
Memory leak fix

### LINKED ISSUE #956: Memory usage
We're repeatedly bumping into our memory limit in production.

screen shot 2016-03-11 at 16 03 27

bjfletcher:
Tell Node to *not* be lazy & greedy using up as much memory as it wishes (up to |1.5 GB). Instead, it has to try and be more aggressive & less greedy.

This is on the assumption we're targeting to work well on a 512 MB machine. We will need to monitor response times for the HTTP actions.

Resolves #956

/cc @i-like-robots @matthew-andrews


bjfletcher:
image

The metrics are looking really good with the test app under a similar load as in production. @i-like-robots

matthew-andrews:
looks good, can we rerun the build again to just make sure that it passes?

PR #1019, opened 2016-04-13
Memory leak fix tweak

bjfletcher:
The objective is to further reduce the number of R14 (memory quota exceeded) errors.

This is a tweak on #1017.  The difference being that the memory limit for the Node app is reduced by 9% from 460 MB down to 400 MB.

Metrics for the resulting deployment should be compared with metrics previous deployments.  Records of which are here: https://github.com/Financial-Times/next-stream-page/issues/956#issuecomment-209486677


bjfletcher:
I'll merge this in the morning (Thursday) and monitor the deployment over the day.

i-like-robots:
👍

Issue #1018, opened 2016-04-13
Fix markets data link on companies pages


i-like-robots:

i-like-robots:
Fixed by #1020

PR #1020, opened 2016-04-14
add symbol to market data transform

andygnewman:
cc: @i-like-robots 

Fixes this issue https://github.com/Financial-Times/next-stream-page/issues/1018


server/transforms/markets-data.js
```diff
@@ -31,7 +31,8 @@ function formatMarketsData(item) {
 		timeStamp: item.quote.timeStamp,
 		volumeTraded: formatVolume(item.quote.volume),
 		changePercent: item.quote.change1DayPercent.toFixed(2),
-		changeDirection: changeDirection(item.quote.change1Day)
+		changeDirection: changeDirection(item.quote.change1Day),
+		symbol
```
bjfletcher: Text editor fail? Shouldn't this `symbol` line be indented to indicate that it's a continuation of the previous line?

andygnewman: I don't think so. It's shorthand for symbol: symbol as a new key: value pair to be returned.

i-like-robots: I don't think we need to pass the `symbol` around, it's in the data (`item.basic.symbol`) =]

i-like-robots:
🚢

PR #1021, opened 2016-04-14
Webpackify

i-like-robots:
Switch to using Webpack for all our client bundling needs. Uses latest n.Makefile so we can use ESLint 2 and JSX linting too.

Dev bundles compile in half the time, prod bundles have proper dead-code elimination (556kb vs. 624kb). So many wins!


.eslintrc.js
```diff
@@ -0,0 +1,50 @@
+const config = {
```
matthew-andrews: can be removed +gitignored now you updated this in `n-makefile`or not ?

i-like-robots: Yes! done

matthew-andrews:
looks good to me, bit worried about the failures though

i-like-robots:
Failures look to be standard flakiness of this app but will double check the saucelabs tests.

Issue #1001, opened 2016-04-06
Related tags should use tag facets API


i-like-robots:
Our current implementation of the [related tags](https://github.com/Financial-Times/next-stream-page/blob/master/server/lib/related-tags.js) functionality relies on [a dependency](https://github.com/Financial-Times/next-ft-api-client/blob/master/src/related-tags.js) that duplicates functionality now found (and improved upon, I hope!) by the [Tag Facets API](https://github.com/Financial-Times/next-tag-facets-api).

As well as migrating this module _into_ the equivalent transform we should also switch to use the new API.

PR #1022, opened 2016-04-14
Switch from using relatedTags method of content API to hit facets API

i-like-robots:
Closes https://github.com/Financial-Times/next-stream-page/issues/1001 and also fixes [this](https://financialtimes.slack.com/archives/ft-next-support/p1460630368000816)


andygnewman:
👍

PR #1024, opened 2016-04-14
use raw loader instead of requireText 

i-like-robots:
Fixes [this](https://financialtimes.slack.com/files/dilip.vaghela/F10M8M6US/screen_shot_2016-04-14_at_14.53.19.png)


client/components/related-content/main.js
```diff
@@ -48,7 +48,7 @@ module.exports = {
 
 				new VideoStream(relatedItemsEl, {
 					items: json.items.slice(0, 2),
-					itemTpl: '<li data-trackable="related item">' + requireText('../video-stream/card.html') + '</li>',
+					itemTpl: '<li data-trackable="related item foobar">' + require('../video-stream/card.html') + '</li>',
```
ironsidevsquincy: foobar

ironsidevsquincy:
:+1:

client/components/related-content/main.js
bjfletcher: Ha ha I'm not the only one then ;)

i-like-robots: 😳

PR #1026, opened 2016-04-14
Do not quote unicode references in CSS

i-like-robots:
Zero padded unicode references get proper munged by the sass loader.

screen shot 2016-04-14 at 17 07 37

screen shot 2016-04-14 at 17 15 34


client/components/related-content/main.js
```diff
@@ -48,7 +48,7 @@ module.exports = {
 
 				new VideoStream(relatedItemsEl, {
 					items: json.items.slice(0, 2),
-					itemTpl: '<li data-trackable="related item foobar">' + require('../video-stream/card.html') + '</li>',
```
matthew-andrews: foobar??

i-like-robots: foobar.

(webpack will only re-build on watch when something _actually_ changes so... whoops)

andygnewman:
👍

i-like-robots:
Hmm, this doesn't actually work quite right... requires further investigation.

client/components/related-content/main.js
matthew-andrews: lol ok

PR #1029, opened 2016-04-16
Fix erroring eu referendum page

matthew-andrews:
@callumlocke @i-like-robots @wheresrhys


server/transforms/stickies.js
```diff
@@ -6,7 +6,7 @@ import legacyTagCompat from '../models/legacy-tag-compat';
 import loadList from '../lib/load-list';
 import Poller from 'ft-poller';
 import subheading from '../models/subheading';
-import { logger } from '@financial-times/n-logger';
```
matthew-andrews: In `n-logger` you should import like this:- https://github.com/Financial-Times/n-logger#usage.  Otherwise, `logger` here will be undefined and [this code](https://github.com/Financial-Times/next-stream-page/pull/1029/files#diff-3d97acf1e6dbe2cf1ed723d1e5842ea6R46), designed to handle failing stickies, will throw…  pulling the whole stream page down.

PR #1030, opened 2016-04-18
Featured content mobile slider

sumeetadur:
Requires https://github.com/Financial-Times/next-stream-page/pull/1007 to go live first


client/main.scss
```diff
@@ -26,6 +26,7 @@ $o-buttons-is-silent: false;
 @import "../shared/components/card-open-in-stream/main";
 @import "../shared/components/custom-fragment/main";
 @import "../shared/components/image-container/main";
+@import "../shared/components/interstitial/main";
```
i-like-robots: I don't think the name fits... an interstitial is used elsewhere within our apps to mean a sort of modal window that is open on page load.

shared/components/interstitial/_main.scss
```diff
@@ -0,0 +1,131 @@
+@import '../highlighted-card/main';
+
+.interstitial {
+
+	margin: 20px calc((100% - 100vw) / 2);
```
i-like-robots: Maybe a comment about what this is doing? Is it because you're after full-width, breaking the container?

shared/components/interstitial/_main.scss
```diff
@@ -0,0 +1,131 @@
...
+
+	.highlighted-card__subheading {
+		display: none;
+	}
+
+	@include oGridRespondTo(L) {
+		margin: 20px 0 0 -20px;
+
+		& > div {
+			-webkit-overflow-scrolling: initial;
```
i-like-robots: instead of resetting all of these values, why not wrap the initial declaration inside a `max-width` media query, E.G. `oGridRespondTo($until: 'L')`

shared/components/interstitial/index.js
```diff
@@ -0,0 +1,28 @@
...
+import HighlightedCard from '../highlighted-card';
+import getGridMeasure from '../../lib/get-grid-measure';
+
+export default class Interstitial extends React.Component {
+	constructor(props) {
+		super(props);
+
+		this.item = (item, index) => {
+
+			return <HighlightedCard item={item} key={index} itemNumber={index} imageSizes={{
```
i-like-robots: I think this could be a stateless component

shared/components/stickies/_main.scss
```diff
@@ -1,7 +1,304 @@
...
+
+		.highlighted-card__image {
+			order: 1;
+			display: block;
+		}
+
+		.highlighted-card__subheading {
+			order: 2;
+			margin-top: 0.75em;
+			font-size: 0.8em;
```
i-like-robots: Last time we checked we were using ~22 font sizes on one page! Could any of these be switched to use the standard set of o-typography sizes?

shared/components/stickies/_main.scss
```diff
@@ -1,7 +1,304 @@
...
+		}
+
+		.highlighted-card__headline { margin-top: -0.15em; }
+
+		.highlighted-card {
+			padding-bottom: 0;
+			padding-top: 0;
+			margin-bottom: 20px;
+
+			&--is-video {
```
i-like-robots: Always stick to writing the class name in full, otherwise we'll never find them!

shared/components/stickies/_main.scss
```diff
@@ -1,7 +1,304 @@
...
+		padding-bottom: 15px;
+	}
+
+	.stickies__others {
+		@include oGridColspan(5);
+
+		.highlighted-card {
+			padding-bottom: 15px;
+
+			&:not(.highlighted-card--2, .highlighted-card--4) {
```
i-like-robots: Does this get transpiled by postcss? AFAIK only a single simple selector is supported.

shared/components/stickies/_main.scss
```diff
@@ -1,7 +1,304 @@
...
+				padding-top: 15px;
+				border-top: 1px solid getColor('pink-tint2');
+			}
+		}
+
+		.highlighted-card__subheading {
```
i-like-robots: `n-util-visually-hidden`

i-like-robots:
Couple of general comments: Is it possible to try and use the standard font sizes where possible? With all the flex containers, how does this look in IE9? I'm really surprised so much CSS was required - I guess there more to it than your screenshot shows?

andygnewman:
@sumeetadur Is this PR still required or can it be closed.

PR #1034, opened 2016-04-20
Enhanced AI & Robotics stream (Robot Week)

sumeetadur:
Preparation for adding custom fragments and stickies components. Uses the "robotWeek" flag.

screen shot 2016-04-20 at 17 38 54


sumeetadur:
@matthew-andrews / @i-like-robots could you have a look through this?

server/transforms/custom-fragments.js
```diff
@@ -22,10 +22,13 @@ export default () => {
...
 		}
 
-		post(res) {
+		post(res, flags) {
 			if (!this.fragments) return;
 
-			for (const { index, html, bleed, title } of this.fragments) {
+			for (const { flag, index, html, bleed, title } of this.fragments) {
+				
+				if(flag && !flags[flag]) return;
```
andygnewman: should make robust incase flags down

```
if(flag && (!flags || !flags[flag])) return;
```

matthew-andrews: if flags are down apps won't start so quite safe to do

`if (flags[flag]) {` 😄

server/transforms/custom-fragments.js
```diff
@@ -2,7 +2,7 @@ import CustomFragmentsPoller from '../lib/custom-fragments-poller';
 import logger from '@financial-times/n-logger';
 
 const poller = new CustomFragmentsPoller(
-	'https://bertha.ig.ft.com/view/publish/gss/13N3rLTt_mcGJQA2aXEXgeVd0J0JIeqoarR2CGnIdnio/manifest'
+	process.env.CUSTOM_FRAGMENTS_MANIFEST || 'https://bertha.ig.ft.com/view/publish/gss/13N3rLTt_mcGJQA2aXEXgeVd0J0JIeqoarR2CGnIdnio/manifest'
```
andygnewman: Is the intention to put the berth a value into config vars for production?

PR #1038, opened 2016-04-27
applying cache and vary helpers from n-express

wheresrhys:
@ironsidevsquincy @andygnewman


server/lib/render-stream.js
```diff
@@ -17,6 +16,12 @@ module.exports = function(req, res) {
 			'application/rss+xml'
 		]);
 
+		// As these are different urls to the main html response we can safely unvary on the things
+		// used by wrapper without worrying about damaging the cache for the canonical url
+		if (req.params.extension === 'rss' || req.query.format === 'rss' || req.params.extension === 'json' || req.query.format === 'json') {
```
bjfletcher: `[req.params.extension, req.query.format].find(x => x.match(/^(rss|json)$/)`
maybe not as readable...

test/server/transforms/live-stream.test.js
```diff
@@ -86,6 +94,10 @@ describe('Live stream transform', () => {
 				expect(res.templateData.isAlive).to.undefined;
 			});
 
+			it('sets correct cache header', () => {
+				expect(res.cache.calledWith('short', {'max-age': 60})).to.be.true;
```
bjfletcher: how about a test that the header has been set?

wheresrhys: part of the reason for using the `.cache` helper is to allow us to easily tweak cache settings across all our apps, so the exact value it sets in the header may change over time, so it seems more sensible to just verify the helper has been called

server/lib/render-stream.js
wheresrhys: There's a comment a few lines down saying using `req.params.extension` will be removed soon, so I was just trying to do the simplest thing for now. Happy to change it if you like though

PR #1041, opened 2016-05-03
Removed code for FT Labs TLDR Feature

seanmtracey:


andygnewman:
I think @ironsidevsquincy is sorting out the next-sass-setup conflicts, but otherwise 👍  assuming it passes the test suite (presume you've run that locally).

ironsidevsquincy:
Yep, on it

seanmtracey:
Ah, OK. I've also resolved the bower conflicts (so I think) and didn't see this message until the network came back up. If I have the wrong versions I'll rollback my last commit

PR #1044, opened 2016-05-05
Hide filters heading and trending filters toggle button for core exp

i-like-robots:


shared/components/stream-filters/_main.scss
```diff
@@ -188,10 +191,15 @@
...
 		margin-left: 0.5em;
+		transition: 0.2s transform 0.25s;
 
 		.is-open + & {
 			// Don't load another icon
 			transform: rotate(180deg);
 		}
 	}
+	
+	.core & {
```
bjfletcher: how does `.core` get added to `.stream-filters__toggle-trending`?

i-like-robots: The classes `core` and `enhanced` are toggled on the document element by the cuts the mustard test

PR #1048, opened 2016-05-09
add Alphaville to custom branded stream pages

andygnewman:
Alphaville now has data in the brand API.

screen shot 2016-05-09 at 12 44 13


matthew-andrews:
oh i didn't realise this was manual, what's stopping us from querying capi for all brands?

andygnewman:
Nothing particularly stopping us - whilst we only had it for a small proportion of the brands it just seemed more efficient to limit it, however it's probably time to readdress this. How many brands do we have with taxonomy of brand?

matthew-andrews:
not that many
http://next-es-interface.ft.com/brand

PR #1049, opened 2016-05-09
Init video when showing less

tom-parker:
@andygnewman @i-like-robots 
 
I think this makes sense. The video component was not getting initialised when going from the open to closed state, so no video was showing.


andygnewman:
LGTM, were you also intentionally upgrading n.Makefile up to v1?

tom-parker:
Yes, so there was a bug that required the latest webpack config which required the latest n.Makefile!

Issue #1050, opened 2016-05-09
Remove custom list of brands with custom headers


andygnewman:
During development of custom header for brands, a config list of brands with these headers was created.

https://github.com/Financial-Times/next-stream-page/blob/master/server/config/data/brands.json

Now it is established, this can be removed and calls to concordance / brands api would be made for all streams in taxonomy of brands.

http://next-es-interface.ft.com/brand.

PR #1051, opened 2016-05-09
remove brands data config and use in profile check

andygnewman:
To close this issue

https://github.com/Financial-Times/next-stream-page/issues/1050

Has consequence of removing related tags (in the News topics) for branded streams that do not have custom header set up.


matthew-andrews:
woah that was quick :]

i-like-robots:
🚢

PR #1052, opened 2016-05-09
Use oAds v4

adgad:
/cc @andrewgeorgiou1981


andrewgeorgiou1981:
👍  (seems to be a flaky test on circleci)

PR #1055, opened 2016-05-10
Steam / Sections - related topics data and code clean up

andygnewman:
Business Requirements:
- Topics with most Articles written about them within current stream.
- Articles under each topic should be only those also tagged against the current stream (ie. intersection)
- Should be flexible as to the maximum number of topics and articles per topic returned.
- Has a minimum number of topics parameter to help manage display (eg. only return if 2 or more)
- If topic has fewer articles than desired number, it should still be returned.

Removed from previous implementation:
- Brand classification (which previously replaced timestamp when branded, eg. Lex)
- Prioritisation of opinion / analysis articles at the top of the list of articles.


server/lib/related-topics.js
```diff
@@ -0,0 +1,64 @@
...
+	});
+
+	let qualifiedTopics = Object.keys(distinctTopics)
+		// Sort topics by most articles and, in a tie, by most recent article
+		.sort((a, b) => {
+			if (distinctTopics[a].items.length < distinctTopics[b].items.length) return 1;
+			if (distinctTopics[a].items.length > distinctTopics[b].items.length) return -1;
+
+			const dateA = new Date(distinctTopics[a].items[0].publishedDate).getTime();
+			const dateB = new Date(distinctTopics[b].items[0].publishedDate).getTime();
```
i-like-robots: `Date.parse()`

server/lib/related-topics.js
```diff
@@ -0,0 +1,64 @@
+'use strict';
+
+const legacyTagCompat = require('../models/legacy-tag-compat');
+
+module.exports = (items, parentConceptId, options) => {
+
+	let relatedTopics;
```
i-like-robots: Maybe make this an empty array

server/lib/related-topics.js
```diff
@@ -0,0 +1,64 @@
...
+					mainImage = item.mainImage;
+					return true;
+				}
+			});
+			topic.mainImage = mainImage;
+			// Reduce number of articles
+			topic.items = topic.items.slice(0, options.articlesPerTopic);
+			topic.trackable = 'stream-hot-topic';
+			topic.template = 'n-topic/templates/card';
+		});
```
i-like-robots: And push these onto it

server/transforms/related-topics.js
```diff
@@ -1,9 +1,12 @@
 'use strict';
 
-const StreamHotTopics = require('../lib/stream-hot-topics');
+import relatedTopics from '../lib/related-topics';
```
i-like-robots: I'd avoid the ES6 module syntax until the implementation of it is agreed

andygnewman:
One last thought on this - does the ordering of the transforms have an impact;
- related topics sets it to get 100 articles on which to assess the topics on pre and then cuts it back down to 50 on post.
- if this happens before opinion / analysis, I presume it limits the pool of articles back to the 50.
- we could swap them around so opinion / analysis has a pool of 100?

i-like-robots:
> we could swap them around so opinion / analysis has a pool of 100

Hmm... yeah. I think it's likely we will continue to pluck 100 stories but ditch pagination for these section pages so may become a non-issue.

PR #1059, opened 2016-05-11
use fetch if the browser has it

TheoLeanse:
@i-like-robots @wheresrhys 

This is OK now, with those Access-Control changes in `next-tag-facets-api`


wheresrhys:
wallop :+1:

PR #1060, opened 2016-05-11
Special Reports performance

### LINKED ISSUE #1032: Improve performance of Special Reports
The special reports home page regularly times out locally, on CI and in production. We need to look at ways to optimise this Elasticsearch query: https://github.com/Financial-Times/next-stream-page/blob/master/server/controllers/shared/special-reports.js

i-like-robots:
1. Sort special report aggregates by date 
2. Only fetch the number of required special reports for the related endpoint
3. Only request the required fields for the top hits aggregation

This isn't a permanent fix, unfortunately paging of aggregations is not possible so we'll continue to request _all_ facets as long so long as we support pagination on this page.

Fixes #1032


matthew-andrews:
:)

PR #1061, opened 2016-05-11
Split out hot opinion/analysis from hot stuff transform

i-like-robots:
- Now a standalone transform
- 👀  to the future


andygnewman:
Suggest that you merge this one before I merge [Related Topics Clean Up](https://github.com/Financial-Times/next-stream-page/pull/1055/files), as you've put in the split out that I'll rebase onto.

Fair play for your patience in making the fixture simpler!

PR #1065, opened 2016-05-16
Remove Robot Week flag and custom fragment CSS

sumeetadur:
Removes the custom fragment CSS and the dependency on the robotWeek flag


andygnewman:
Would it be better to remove the full profile for this page rather than just remove the specific robot transform?

sumeetadur:
We just had a discussion about this. If we remove the full profile then the relatedTags transform is added through the default profile which we don't want. Keeping the profile with just the default transforms is the easiest way to have this right now but we need to discuss how to change this so that any streams with custom fragments or stickies will remove the relatedTags by default.

andygnewman:
We're starting the Stream page redesign work now and the related tags element looks like it will be going, replaced by some form of Nav / Breadcrumb type approach, so when this is in we can change that.

sumeetadur:
Perfect, thanks!

sumeetadur:
@andygnewman can I get a 👍 ?

andygnewman:
👍

PR #1068, opened 2016-05-17
Moves o-expander into the head.css

GlynnPhillips:
o-expander was used above the fold but not included in the head.css

Therefor the presentation of the page was broken if the main.css failed to
download or more importantly was slow to download.


i-like-robots:
There's 200kb+ of CSS in the head already.

TBH I vote we simply switch it off for now rather than shove more in.

GlynnPhillips:
I guess the issue we are facing with critical css on next (and indeed a lot of modern sites) is that the styles required to do 'above the fold' makes us the bulk of styles due to repetition of patterns / grids. So we probably need some numbers for how much is too much in the head, I know its difficult because its always a 'it depends' answer but else we will continue having these discussions.

phamann:
> There's 200kb+ of CSS in the head already.

I'm assuming that is uncompressed (pre-GZip)?

> I guess ideally we would have some vague numbers for how much is too much in the head

Ideally, you want 16Kb gzipped, to fit inside the initial TCP congestion window. But for most sites this ins't achievable. @ironsidevsquincy has been aiming for 20-30kb per page, which I think is perfectly acceptable.

> TBH I vote we simply switch it off for now rather than shove more in.

Switch off what? The inline CSS or the feature? As the streams are being rebuilt I think its safe to jsut switch off the feature.

My two pence about above-the-fold debate, is if a feature isn't critical for it to be loaded in head CSS, why is it even in the above-the-fold area of the page anyway? It's also OK to only show non-critical features with JS after load a,k.a lazy loading

wheresrhys:
I say merge this fix. Very soon we'll be able to roll out n-ui v1 (I'm just test-driving in article) which'll make it a lot easier to manage this as the js can hook onto a mainCss load promise.

wheresrhys:
ps expander css is tiny, so isn't really the cause of the bloat

bjfletcher:
LGTM :+1:

GlynnPhillips:
`head.css` gzip'ed on master 26k
`head.css` gzip'ed on this branch 27k

I am going to merge for now so the reported bug is fixed and then we can discuss this.

PR #1069, opened 2016-05-17
IE9 doesn't support dataset

adgad:
This was causing JS to break on special report stream pages.

/cc @andygnewman @bjfletcher @i-like-robots


i-like-robots:
🚢

PR #1070, opened 2016-05-17
Fixes the presentation and loading of related articles

GlynnPhillips:
The loading of the related articles was broken due to a url change and
the presentation was also wrong.

@andygnewman


andygnewman:
You could add a smoke test for the url used in the client side javascript, so we know if we break it again.

GlynnPhillips:
@andygnewman It wasn't actually the url in the client side that broken instead it was a match on the current page url on the server side that broke and it has now been altered so it doesn't match on the url anymore.

It used to be this: https://github.com/Financial-Times/next-stream-page/blob/master/server/config/profiles.js#L147

Now its this: https://github.com/Financial-Times/next-stream-page/blob/f1ca4ad0891bd5d755cdd3123fbf64e9696ae194/server/config/profiles.js#L147

andygnewman:
Sorry, of course we went through that when we sat down.

PR #1071, opened 2016-05-18
Fixes responsiveness of related topics

GlynnPhillips:
The modifier class was not using the correct property to check the
length so on smaller screens the cards were getting very small.


andygnewman:
Good spot, I must have missed that when I refactored the back end of it the other day.

PR #1072, opened 2016-05-18
As an ID can appear in multiple places in the nav only use the matchi…

matthew-andrews:
…ng nav item if we haven't found matching one already or if the one that has been matched has children.

From:-
screen shot 2016-05-18 at 12 23 53

To:-
screen shot 2016-05-18 at 12 23 38

e.g. https://next.ft.com/companies/financials


bjfletcher:
LGTM

PR #1073, opened 2016-05-18
Use homepage components

bjfletcher:
First pass at getting the stream app to use the `n-section` components (which are the components that homepage uses).

Next pass is to setup data and plumb it into the aforementioned components.


server/setup.js
```diff
@@ -1,5 +1,5 @@
 // modules that need to be compiled by babel
-const es6Modules = ['next-myft-ui', 'n-image'];
+const includes = require('../n-makefile.json').assets.includes;
```
andygnewman: I would suggest a naming change here - the includes is confusing where you have filename.includes and !includes.some on the same line.

views/article-list.html
```diff
@@ -1,3 +1,9 @@
+{{#if flags.streamNewPages}}
+<h1>v N-SECTION COMPONENT v</h1>
+{{{reactRenderToString Section spread=sectionData}}}
```
andygnewman: What's the behaviour of this at this point in time?
I'd rather not promote stuff, even though it is behind a flag, that just results in 500's, which i suspect is what happens here with the flag on?

server/setup.js
```diff
@@ -14,7 +14,16 @@ require('babel-register')({
 		// https://github.com/59naga/babel-plugin-add-module-exports
 		'add-module-exports',
 	],
-	ignore: filename => (
-		filename.includes('/node_modules/') && !es6Modules.some(module => filename.includes(`/node_modules/${module}`))
-	),
+	ignore: filename => {
```
andygnewman: The conversation in slack around this seemed to pass me by, does this reflect the wisdom of the crowd?

views/article-list.html
bjfletcher: @andygnewman I don't see any 500's with this branch.

There's a lot of 500's with the `master` branch though.

server/setup.js
bjfletcher: :+1:

views/article-list.html
bjfletcher: @andygnewman I'm happy to remove that block from the template, if you like. It's really just a way of testing that the `n-section` component gets rendered ok when the flag is turned on.

server/setup.js
bjfletcher: Yes. It's enough to look in the specific modules (listed in `n-makefile.json`) and anywhere outside `node_modules` and `bower_components`.  In the future, if we find that we need JS code from `bower_components`, we can add to the code above.

ironsidevsquincy: what's the benefit over this over how it was before? (just asking to know if we should do this elsewhere)

don't think we'd ever use bower components on the server side

andygnewman:
LGTM, 👍

server/setup.js
bjfletcher: @ironsidevsquincy it feels right to have the list of modules in `n-makefile.json`

(we may want to split that `includes` one up into two, one for babel-register - server - and one for webpack's babel - client)

I think it is clearer and more readable than what it was before, but maybe not and it's just me :) only difference is to explicitly exclude `bower_components` and that if we find that we need something in `bower_components`, we see this and realise we'd need to include it

even if `bower_components` is already excluded by default, I think it's nice to be clear, sure and explicit in here

ironsidevsquincy: no, that makes sense, `n-makfile` now contains all external deps that need babel-ifying

I don't think it needs to be this verbose though. there is no default 'ignore' now that we're explicitly setting this ignore flag, and we shouldn't be including bower components on the server, so shouldn't even have that option here

ironsidevsquincy: sorry, mean this function can be as verbose as you like, just don't think there should be that `} else if (filename.includes('/bower_components/')) {` condition

PR #1076, opened 2016-05-19
Fix /news-feed?format=json URLs

matthew-andrews:


server/controllers/stream.js
```diff
@@ -96,6 +97,7 @@ module.exports = (req, res) => {
 
 	// Or fail gracefully
 	.catch(err => {
+		logger.error(err);
```
matthew-andrews: this would help with visibility :]

bjfletcher:
:=1: for the log, let's merge this

PR #1078, opened 2016-05-19
enable the videoJS player behind a flag

VladDubrovskis:
/cc @leggsimon @adgad @andrewgeorgiou1981


client/components/related-content/main.js
```diff
@@ -45,10 +45,18 @@ module.exports = {
 
 				fetchedVideos = json.items;
 
+				let videoPlayer = '';
+				if(flags.get('brightcovePlayer')) {
+					videoPlayer = 'brightcovePlayer';
```
matthew-andrews: i think we can probably delete this

VladDubrovskis: I think there was one for brightcovePlayer in n-section or front page. will remove from there too

matthew-andrews: thank you for taking the time to tidy this up.

VladDubrovskis: No worries, Seems like i might have imagined seeing it elsewhere. Searched github and local repos and this is only place.

Shall I delete the flag from from flags api?

matthew-andrews: :) yes please

PR #1079, opened 2016-05-19
Get stream app running locally

matthew-andrews:


bjfletcher:
to be clear, this is for the news-feed page... there may still be some other modules we'll need to include in the list

matthew-andrews:
step by step :]

PR #1081, opened 2016-05-19
Mock data & layout using n-section

bjfletcher:
Captured and took top stories layout and data from frontpage to here, to experiment with.


server/transforms/trending-topics.js
```diff
@@ -0,0 +1,22 @@
...
+	return class {
+		pre (req) {
+			this.isFirstPage = req.perPage.page === 1;
+		}
+
+		post (res, flags) {
+			if (this.isFirstPage && flags.streamNewPages) {
+				const items = res.templateData.stream.items;
+				const conceptId = res.templateData.conceptId;
+				const data = { content: relatedTopics(items, conceptId, { maxTopics: 3, minTopics: 3, articlesPerTopic: 2 }).map(t => ({ type: 'Concept', id: t.term.id, name: t.term.name, url: t.term.url, taxonomy: t.term.taxonomy, items: t.items.map(i => ({ id: i.id, title: i.title, isPodcast: false, primaryImage: { rawSrc: t.mainImage.url }})) })) };
```
bjfletcher: I do a lot of `.map` operations here mapping the `relatedTopics` data structure to the `n-section`/`n-card` data structure.  It's working fine, I see the latest trending topics in the view.

I wonder if I should:

a) keep that (of course move that long line into a idented function of its own in the module)
b) bypass `relatedTopics` call and map from `items` directly instead
c) anything else you think I should do instead

@andygnewman @i-like-robots

andygnewman: I wouldn't introduce a trendingTopics transform that basically wraps relatedTopics when relatedTopics already does the job and will in effect be replaced by moving the display of it.

I would instead use the flag mechanic to alter the behaviour in relatedTopics.

The key question you have is about how to transform the data to make n-section work with it.
This is an issue I want to grapple with next week to abstract the article transformation that sits within our GraphQL API (and is aligned to n-section), out of there and into a module that can be used more widely.

Until I've done this, I would recommend you continue to use the type of method you have above.

server/transforms/top-stories.js
```diff
@@ -0,0 +1,27 @@
...
+const fetchGraphqlData = require('../lib/fetch-graphql-data');
+
+module.exports = () => {
+
+	return class {
+		pre (req) {
+			this.isFirstPage = req.perPage.page === 1;
+		}
+
+		post (res, flags) {
```
andygnewman: Would recommend you switch this to happen on pre rather than post (this was the feedback I had from @i-like-robots on most read and applies here too)

The stream id is available on the req object.

Have changed my most read one to a better pattern, but not yet merged ...https://github.com/Financial-Times/next-stream-page/pull/1086/files

andygnewman: Should have added this wiki link also. https://github.com/Financial-Times/next-stream-page/wiki/Transforms

bjfletcher: @andygnewman Great. I'll copy over your new code when the real data is ready.

bjfletcher:
@andygnewman it's ready to go in.  A few notes:

### Layout config
- see https://github.com/Financial-Times/n-section/pull/40 which will mean our config helpers disappearing soon in favour of importing from `n-section`

### Data
- top stories are using mock data, we will work on the transforms to use actual data in a separate branch
- trending topics use a mapping hack to reshape the data, we will work on a better approach in a separate branch

server/config/profiles.js
```diff
@@ -89,6 +91,8 @@ module.exports = [
 			relatedTopics(),
 			currentNav(),
 			relatedTags(),
+			topStories(),
```
andygnewman: If we're just using mock data and having stuff behind the flag, we may not want to put the transform into all the places yet?

bjfletcher: @andygnewman The server don't NEED to do this, I agree.  Three reasons I don't put all that behind a flag:

1. profiles don't have access to flags (yet) and I don't think it's worth the effort;
2. user won't see anything

If you're particularly keen to work on this, I think that's for a separate branch. This branch is already very large as it is.

andygnewman:
👍

bjfletcher:
Thanks @andygnewman :)

PR #1082, opened 2016-05-20
Most Read on Stream Pages

andygnewman:
Current State

screen shot 2016-05-20 at 18 14 02

To Do
- [x] Add graphql API key into config vars for stream app
- [x] Make request to graphql use the conceptid variable (need to merge this [PR](https://github.com/Financial-Times/next-graphql-api/pull/137) first
- [x] Styling appropriate to background color
- [x] Write tests

Done
- [x] add most read by concept id poller & end point to most read api
- [x] add config to graphql to get this end point and fetch the content
- [x] stream transform for most read, including querying graphql and decorating object in format required for n-section (this really needs abstraction from `next-front-page`, but so far have just copied it.


Makefile
```diff
@@ -7,13 +7,13 @@ coverage-report:
 	nyc --reporter=$(if $(CIRCLECI),lcovonly,lcov) make unit-test
 
 unit-test:
-	export AWS_SIGNED_FETCH_DISABLE_DNS_RESOLUTION=true; export apikey=12345; export api2key=67890; export HOSTEDGRAPHITE_APIKEY=123; export ENVIRONMENT=production; mocha --require server/setup 'test/server/**/*.test.js'
+	export AWS_SIGNED_FETCH_DISABLE_DNS_RESOLUTION=true; export apikey=12345; export api2key=67890; export HOSTEDGRAPHITE_APIKEY=123; export ENVIRONMENT=production; mocha --require server/setup 'test/server/**/*.test.js' 'test/config/**/*.test.js'
```
i-like-robots: I think the glob can be `test/(server|config)/**/*.test.js`

Makefile
```diff
@@ -7,13 +7,13 @@ coverage-report:
...
 
 component-specs:
 	mocha --require server/setup 'test/shared/components/**/*.test.js'
 
 test:
-	make verify unit-test component-specs
```
bjfletcher: just checking this `unit-test` removal was intentional, if it was, how'd CircleCI run the unit tests?

package.json
```diff
@@ -18,6 +18,7 @@
     "@financial-times/n-raven": "^2.0.0",
     "@financial-times/n-section": "^3.1.5",
     "babel-plugin-add-module-exports": "^0.1.2",
+    "babel-plugin-array-includes": "^2.0.2",
```
i-like-robots: Just use `.indexOf()` or `.find()` or other suitable array method?

server/transforms/most-read.js
```diff
@@ -0,0 +1,30 @@
...
+
+	return class {
+		pre(req) {
+			this.isFirstPage = req.perPage.page === 1;
+		}
+
+		post(res, flags) {
+			if (this.isFirstPage && flags.streamNewPages) {
+				const conceptId = res.templateData.conceptId;
+				return fetchGraphqlData(mostReadQuery, {concept: conceptId})
```
i-like-robots: Worth considering the non-blocking async way? https://github.com/Financial-Times/next-stream-page/wiki/Transforms

Makefile
andygnewman: The lines from 18 - 22 below ensure that unit-test is run either as part of the test coverage report, or on it's own.
Unit tests were running twice otherwise.

package.json
andygnewman: Yes, hated doing this, but unfortunately the includes is in `n-section`, I could do a PR on that to get rid of.

server/lib/fetch-graphql-data.js
```diff
@@ -0,0 +1,22 @@
...
+		body: JSON.stringify({ query, variables }),
+		headers: {
+			'Content-Type': 'application/json'
+		},
+		timeout: 3000
+	})
+		.then(fetchres.json)
+		.catch(error => {
+			logger.warn('Failed getting graph ql api response', {
+				error: error
```
bjfletcher: Great... maybe include `query` & `variables` in the error log? It'd be good to know whether the error was because of the 3000ms timeout, but don't worry if not sure. We can cross this bridge later.

package.json
matthew-andrews: probably not helpful but is the polyfill service not an option?

client/components/top-section/_main.scss
```diff
@@ -0,0 +1,15 @@
...
+	}
+	.section--most-popular {
+		margin-top: 20px;
+		margin-bottom: 20px;
+	}
+	.section__column__inner {
+		padding-bottom: 20px;
+		border-bottom: 1px solid getColor('warm-3');
+	}
+}
```
bjfletcher: I'm confused with this `.scss` file. Seems random. What is a `top-section`? Perhaps rename file to `section.scss` or something? Just letting you know it's not clear to me here what this file is for. Maybe a chat on Slack to get me up to speed on all this?

bjfletcher:
Looking really, really good. :) Very much looking forward to merging this and seeing it live on our site.

package.json
bjfletcher: Matt, interesting... how'd I use polyfill.io in Node apps?

Makefile
andygnewman: Close `'test/{server,config}/**/*.test.js'`

client/components/top-section/_main.scss
andygnewman: Good point - the intention of this was to have a single home for overrides to styling from n-section (with the intention that it could possibly get ported back into n-section).

So I agree that section is a better name for it.

server/lib/fetch-graphql-data.js
andygnewman: Yes, I was hesitant to put in the full query given it's size, but variables would indicate the stream id, so I'll see if there's also a property in the query that indicates the type of query it is.

server/transforms/most-read.js
andygnewman: Yes good idea, will do in a separate PR after deploying this one.

bjfletcher:
🎉

PR #1087, opened 2016-05-25
 add billboard ad size to the top slot. and add the the adLoadingImpr…

VladDubrovskis:
…ovement behind flag
/cc @adgad @andrewgeorgiou1981


adgad:
👍

PR #1090, opened 2016-05-26
Section breadcrumbs

i-like-robots:
This allows the current navigation transform to keep track of the branches it follows allowing us to build a breadcrumb.

related: https://github.com/Financial-Times/n-header-footer/pull/211


server/config/profiles.js
```diff
@@ -31,6 +31,7 @@ const defaultTransforms = [
 	customFragments(),
 	stickies(),
 	featuredContent(),
+	currentNav()
```
bjfletcher: is the idea to have the breadcrumbs on *every* page?

i-like-robots: Every page that is in the navigation structure defined in https://github.com/Financial-Times/n-navigation

bjfletcher: image

reason I'm asking about `currentNav` is that with this included, the "related" view goes in the view. I have no idea what the relationship between those two are, but if `currentNav` is included by default, we'd want to look at the "related" view too. :)

andygnewman: In it's current rendering in the subhead there were some good reasons not to have it as a default transform, eg. brands subhead (eg. Westminster Blog), and some of the Brexit style stuff with custom headers.

Agree that when styled horizontally across the top that it should be a default.

i-like-robots: Have rolled back changes to this module

PR #1092, opened 2016-05-26
Add a spec for a profiles config module

i-like-robots:
1. This may ruin the code coverage stats as transform modules missing specs will be instrumented
2. I think this is a sensible check for our own sanity


bjfletcher:
Neato. A big :+1: :)

Out of interest, do you still use `use strict;` at top of JS files?

i-like-robots:
For Node, yes - for proper block scoping of `const` and to allow usage of `let`.This app uses Babel but it stops my editor complaining too.

PR #1093, opened 2016-05-26
Remove section title background colour override

sumeetadur:
Let the section titles inherit the FT pink background colour from n-section


sumeetadur:
@i-like-robots / @matthew-andrews could you peer review this please?

andygnewman:
👍

PR #1094, opened 2016-05-27
change to use serif instead of sans for the heading and better align the follow button

bjfletcher:
Go from:
image

to:

image

Waiting for confirmation from UX that that's what they want.

/cc @andygnewman @i-like-robots


client/components/sub-header/_main.scss
```diff
@@ -176,8 +176,8 @@
 }
 
 .sub-header__page-title {
-	@include oTypographySansData(xl, $load-progressively: true);
-	@include oTypographyFallbackFontSize(sansData, 40px);
+	@include oTypographySerif(xl, $load-progressively: true);
```
andygnewman: I think this should be SerifDisplay https://github.com/Financial-Times/o-typography#serif

client/components/sub-header/_main.scss
```diff
@@ -176,8 +176,8 @@
 }
 
 .sub-header__page-title {
-	@include oTypographySansData(xl, $load-progressively: true);
-	@include oTypographyFallbackFontSize(sansData, 40px);
+	@include oTypographySerifDisplay(xl, $load-progressively: true);
+	@include oTypographyFallbackFontSize(serifDisplay, 40px);
```
bjfletcher: @andygnewman I've changed to SerifDisplay. I'm not sure the fallback font will work. Do you know?

andygnewman: Should do https://github.com/Financial-Times/o-typography/blob/master/scss/_progressive-font.scss#L42

bjfletcher: Great, thanks Andy :)

PR #1104, opened 2016-05-31
Add current nav transform to brand streams

i-like-robots:


andygnewman:
Why? Doesn't this impact the custom header.

i-like-robots:
> Why?

Lex, FT View etc. are in the navigation hierarchy 

> Doesn't this impact the custom header. 

I don't think so - the breadcrumb is standalone and unrelated to the stream header.

andygnewman:
Just tested it out.

The only weirdness I can find is a brand, which has description (but no image) gets the related sections in the sub header. Like this Lex one.


screen shot 2016-05-31 at 17 32 14

i-like-robots:
I think that's acceptable... as this all in flux but I guess it's not up to me =]

andygnewman:
Yep, I think you're right. The sooner it just becomes the breadcrumb, the better.

i-like-robots:
Sub sections now off when breadcrumb is on

andygnewman:
LGTM

PR #1108, opened 2016-06-01
Toggle myFT save buttons with myFtApiWrite flag

leggsimon:
cc/ @laurieboyes @keirog @andygnewman


keirog:
I think now is a good time to get rid of the `saveForLater` flag

keirog:
But otherwise, totally 👍

n-makefile.json
```diff
@@ -9,6 +9,7 @@
       "@financial-times/n-card",
       "@financial-times/n-section",
       "@financial-times/n-image",
+      "@financial-times/n-ui",
       "next-myft-ui",
```
laurieboyes: We potentially don't need `next-myft-ui` this now that it's contained within `n-ui`

PR #1112, opened 2016-06-02
Use real data for Top Stories

bjfletcher:
Use APIs, instead of mock data, and Andy's model transformation.


bjfletcher:
With this first attempt, I'm getting this error:

`level=error event=uncaughterror error_message="Cannot read property 'concat' of undefined" error_name=TypeError error_stack="TypeError: Cannot read property 'concat' of undefined`

it looks like the data from `n-content-model` isn't in the expected format for `n-section`. /cc @andygnewman

server/transforms/top-stories.js
```diff
@@ -1,20 +1,98 @@
...
-				// hack data to show opinion/analysis instead
-				const opinionAndAnalysisData = Object.assign({}, data, { content: data.content.map(i => Object.assign({}, i, { tags: i.tags.map(t => Object.assign({}, t, { name: 'Comment' })) })) });
-				res.templateData.opinionAndAnalysis = getSection('opinion-and-analysis', opinionAndAnalysisData);
+				return this.listFetch.then(summaries => {
+					res.templateData.leadStory = getSection('lead-story', {
+						content: summaries.slice(0, 1)
+					});
+					const summaryIds = summaries.map(summary => summary.id);
+					// find opinion/analysis items in stream
+					const opinionAndAnalysis = res.templateData.stream.items.filter(item => item.metadata && itemGenreIs(item, commentTags))
```
andygnewman: When you get on to doing the Opinion and Analysis piece for real it may be worth reutilising this https://github.com/Financial-Times/next-stream-page/blob/master/server/transforms/opinion-analysis.js, depending on the approach.

andygnewman:
Some nice tests, have you run the coverage tests locally? they give some great clues as to where tests are missing.

server/transforms/top-stories.js
bjfletcher: :+1:

bjfletcher:
@andygnewman around 90% coverage, will try get it to 100% in the PR with the final logic

PR #1118, opened 2016-06-06
Put in the new page the old stream list, so that we can ship

bjfletcher:
... the new Top Stories and Trending Topics.


andygnewman:
I think that there is too much duplication in here.

I would have structured it along the lines of;

if streamNewPage {
    topStories
    opinionAndAnalysis
    trendingTopics
}

unless streamNewPage {
    hotComment
    hotTopics
}

if streamMostRead {
    mostRead
}

streamList

Let me know if that doesn't make sense.

PR #1121, opened 2016-06-07
More curated top story lists

bjfletcher:
See https://docs.google.com/spreadsheets/d/15OvBbIecQ5RUiHCuUpzYnKd7UkWIF15W3yw3q-s_FRE/edit#gid=0

Please note that I put Life and Arts in along with the other top-level sections.

/cc @andygnewman could you please review? Thanks


bjfletcher:
i tested Life and Arts with flag on and off, seemed ok.

andygnewman:
Would this make Hot Opinion Analysis and Hot Topics appear on Life & Arts / Business Education when the streamNewPage flag is off?

Happy for those elements to appear on US page when flag is off.

andygnewman:
I wonder whether those components not appearing with the flag off is more accidental /coincidental than deliberate - ie. the reason they don't appear is that there are no trending topics and no opinion and analysis pieces. Doesn't always mean that it will be that way.

Would rather that they have their own transform group that doesn't expose this possibility.

bjfletcher:
@andygnewman just checked with Max... he thought it was ok to show the Hot OA & Hot Topics in the old pages for these:

Life & Arts
Business Education
Personal Finance

for now.  We :boom: them when we ship the new stuff :-)

PR #1122, opened 2016-06-07
Remove related tags from stream headers

andygnewman:
Removes from all stream pages, regardless of whether they have / will have trending topics.

screen shot 2016-06-07 at 17 07 06
screen shot 2016-06-07 at 17 06 47
screen shot 2016-06-07 at 17 06 18
 

Will still need work to clean up stream headers, which I would propose gets done along with the author / brand / company piece.


bjfletcher:
Woohoo! 💥 
:+1:

PR #1123, opened 2016-06-08
Curated opinion & analysis

bjfletcher:
This completes the Opinion & Analysis work.  Including these cards:

Sections: Add curated O&A lists and fallback to automated if empty
Sections: Hide O&A header if that column is populated by Top Stories

Test coverage is also now 100%.


server/transforms/top-stories.js
```diff
@@ -34,13 +43,15 @@ module.exports = () => {
...
-				const listUuid = listUuids[conceptId];
-				this.listFetch = api.pages({ uuid: listUuid })
+
+				// get curated list of top stories
+				const topStoryListUuid = topStoryListUuids[conceptId];
+				this.topStoryListFetch = api.pages({ uuid: topStoryListUuid })
 					.then(list => list.slice(0, 9))
 					// get top stories from ES using list
-					.then(list => api.content({ uuid: list }))
+					.then(list => list.length ? api.content({ uuid: list }) : [])
```
andygnewman: should the [] at the end be wrapped in a Promise.resolve([])?

server/transforms/top-stories.js
```diff
@@ -62,35 +73,68 @@ module.exports = () => {
...
+							content: topStorySummaries.slice(1, 5)
+						});
+						res.templateData.opinionAndAnalysis = getSection('small-top-stories', {
+							content: topStorySummaries.slice(5, 9)
 						});
+
+					} else if (commentSummaries.length === 0) {
+
+						// use automated opinion & analysis
+						const summaryIds = topStorySummaries.map(summary => summary.id);
```
andygnewman: When filtering automated O&A vs topStories, you would only need to dedupe against the first 5 so could do;
`const summaryIds = topStorySummaries.slice(0,5).map(summary => summary.id);`

server/transforms/top-stories.js
```diff
@@ -62,35 +73,68 @@ module.exports = () => {
...
+
+					// second column
+					if (typeof commentSummaries !== 'object') {
+
+						// no opinion & analysis, use top stories
 						res.templateData.topStories = getSection('top-stories', {
-							content: summaries.slice(1, 5)
+							content: topStorySummaries.slice(1, 5)
+						});
+						res.templateData.opinionAndAnalysis = getSection('small-top-stories', {
```
andygnewman: Not one for right now, but we need to think about the structure of the data here vs presentation - this naming highlights the issue.

andygnewman:
Just a couple of minor comments - good work on the tests!

server/transforms/top-stories.js
bjfletcher: `.then` wraps it if it's not a promise

server/transforms/top-stories.js
bjfletcher: The `slice` for top stories is done earlier (see line 52).

server/transforms/top-stories.js
bjfletcher: Agreed. `opinionAndAnalysisOrTopStories` maybe too long. I want to be descriptive what this is, rather than being too generic like with "second column". Maybe "topSecondColumn"? Nah. Hmm.

server/transforms/top-stories.js
andygnewman: The slice you do above is to take it down to 9. I was suggesting that rather than dedupe against the full 9, as you know that you're only going to use the top 5 you should just dedupe against the top 5.

PR #1124, opened 2016-06-08
Increase trending topics

bjfletcher:
Increased number of links in each trending topic from 2 to 3.

Added a test for the trending topics transform with 100% coverage too.


andygnewman:
👍

PR #1127, opened 2016-06-08
Related / Trending topics enhancements

andygnewman:
Trello Card: https://trello.com/c/Y3QHqRks
- only use articles in last 7 days for related topics function
- bump trending topics transform to above related topics, so gets to pick from upto 100 articles
- enable a minimum number of articles per topic as well as a max (will have min 2 for non-section pages)
- return an empty array if no topics meet all the criteria
- test for new scenarios
- improve tests for related topics transform vs lib so independent


bjfletcher:
Very nice @andygnewman! :+1: :-)

PR #1128, opened 2016-06-08
[WIP] trending topics for all streams

andygnewman:
Second half of this Trello card https://trello.com/c/Y3QHqRks

To Do 
- [x] abstract fetch 100 articles on first page as a new transform to all streams
- [x] add trending topics to all (most) streams => what to do for brexit etc? => leave for now
- [x] add to template for new streamList in RHR if Top Section not present and Trending Topics exist


server/lib/render-stream.js
```diff
@@ -35,19 +36,20 @@ module.exports = function (req, res) {
 			if (res.locals.flags.streamNewPages || res.locals.flags.streamMostRead) {
 				templateData.Section = components.Section;
 			}
-			// show new stream page if flag and all the data it needs is present
+			// show new top section if flag and all the data it needs is present
 			if (res.locals.flags.streamNewPages && ['leadStory', 'topStories', 'opinionAndAnalysis'].every(id => templateData.hasOwnProperty(id))) {
```
bjfletcher: I remember you putting in place `hasTopStories` in https://github.com/Financial-Times/next-stream-page/blob/master/server/transforms/top-stories.js#L143 - what was that for? (Just so I'm aware.)

server/lib/render-stream.js
```diff
@@ -35,19 +36,20 @@ module.exports = function (req, res) {
...
 				templateData.hasStreamListTitle = true;
 			}
+			// use new stream list
 			if (res.locals.flags.streamNewList && !templateData.openInStream) {
 				templateData.streamNewList = true;
-			}
-			if (templateData.streamNewList) {
 				templateData.StreamNew = StreamNew;
 			}
+			// adding 'old' stream react component as default
```
bjfletcher: appreciate the helpful comments

server/transforms/trending-topics.js
```diff
@@ -4,7 +4,7 @@ const getSection = require('../../config/sections/index');
 function toSectionData (items, conceptId) {
 	const relatedTopicsData = relatedTopics(items, conceptId, {
 		maxTopics: 3,
-		minTopics: 1,
+		minTopics: 2,
```
bjfletcher: Why not 3? The new section design asks for 3. Just a thought.

bjfletcher:
:+1: LGTM

server/lib/render-stream.js
andygnewman: Good memory and you're right there's duplication.
Can get rid of the hasTopStories (don't think it is actually being used anywhere) and exclusively use this one (Though do you still want the logging if any of the elements are missing?
Going forwards we will need some indicators of what properties the page has to either control display logic or populate spoor data (eg. hasTrendingTopics, hasTopStories, hasOpinionAndAnalysis, hasMostRead, hasFilters, isLive) we could create a `properties` property of `templateData` where we store all these and run it out of `render-stream`?
Probably best picked up when we get around to the data card.

server/transforms/trending-topics.js
andygnewman: To avoid big empty spaces.

The minimum comes into play if the code can not populate three trending topics. Should not really be a problem on Section Pages (except Life & Arts - will explain that later).

However as this extends trendingTopics to all streams, we want to give the component a chance of appearing on a good number of streams, especially with the 7 day rule implemented. (especially as the design for trending topics on these pages is not so dependent on there being 3).

So trendingTopics is returned if it manages to populate 2 or 3 different topics. If it can only populate 1, an empty array is returned.

The Life & Arts issue is that they write on such a variety of specific topics that each trending topic generally only has one story in it.

We were wanting to implement a 2 article minimum per topic (otherwise it's not really a trending topic). But this would result in a big empty space on the right.

We should get to this in the Weekend Journey work that we're likely to do at the backend of streams, at which point we can replace trending topics on Life & Arts with a different component and put the 2 article per topic minimum in place.

A bit rambling, but hope that explains.

PR #1130, opened 2016-06-09
30 items per page

wheresrhys:
@bjfletcher @andygnewman @mmockett


mmockett:
This should be fine so long as there are no components being injected after the 30th item at the moment. @andygnewman could tell you if so.

andygnewman:
@wheresrhys We have an open question in our work over how many items to display per page.

Was there some analysis or thinking that went into your thinking about 30?

(We do have full data available on the index of which stories get clicked on on stream pages.

andygnewman:
Oh and our other issue with reducing the number of stories is that we actually go and get 100 stories to populate things like related / trending topics, then cut back down to 50. So would need a few more changes to get a consistent proposition with today.

wheresrhys:
Could you point me in the direction of data on what gets clicked and any places I might need to make code changes please.

I am a little bit wary of interpreting 'n% of users clicked on item N' as 'we should keep at least N items' though. We don't know that they wouldn't be fine with either licking a pagination link or clicking on a different headline

andygnewman:
@wheresrhys I'm in the middle of refactoring how we do the fetch number for stories vs display number.

Will point you at it once I've pushed it live.

Are you assuming that there will be a significant performance improvement in only rendering 30 stories?

andygnewman:
This is where we determine the number of articles to fetch (on pre) and cut it back down for rendering (on post).
https://github.com/Financial-Times/next-stream-page/blob/master/server/transforms/fetch-number.js

andygnewman:
It has been cut down to 25, so closing this one.

PR #1131, opened 2016-06-09
update version of o-video to one that contains more ad targeting

VladDubrovskis:
/cc @adgad @lc512k @leggsimon


andrewgeorgiou1981:
👍

PR #1132, opened 2016-06-09
Ad placement

bjfletcher:
Places the ad in the white space underneath Opinion & Analysis (OA) on M and underneath OA on XS and S. It does not show on L and XL.

Tested on Chrome in all breakpoints.


client/components/section/_main.scss
```diff
@@ -58,3 +58,12 @@
 		display: none;
 	}
 }
+
+.advert {
```
andygnewman: We have an ads SCSS file already, may be better to put in there.
https://github.com/Financial-Times/next-stream-page/blob/master/client/components/ads/_main.scss

config/layouts/opinion-and-analysis.js
```diff
@@ -19,6 +19,17 @@ export default [{
...
 			size: 'small'
+		},
+		{
+			type: 'Row',
+			components: [{
+				type: 'Column',
+				colspan: { default: 12, L: 'hide' },
+				components: [{
+					type: 'Section',
+					raw: '<div class="advert o-ads" data-o-ads-name="mid" data-o-ads-targeting="pos=mid;" data-o-ads-center="true" data-o-ads-formats-default="MediumRectangle" data-o-ads-formats-small="MediumRectangle" data-o-ads-formats-medium="MediumRectangle" data-o-ads-formats-large="false" data-o-ads-formats-extra="false" data-o-ads-lazy-load="true" aria-hidden="true"></div>'
```
andygnewman: I'll trust that this is the correct configuration for the ad slot - is there documentation on these configurations anywhere?

bjfletcher: @andygnewman It came straight from Arjun, more information in these links:
https://github.com/financial-times/o-ads
http://registry.origami.ft.com/components/o-ads

so, the ad has the name "mid" whereas the top banner has the name "top".

It's not responsive. If we need it to be responsive, will need to add `responsive` after `MediumRectangle` with a comma in between.

adgad: https://github.com/Financial-Times/n-ui/blob/master/ads/README.md and https://github.com/Financial-Times/o-ads/blob/master/docs/SLOT_CONFIG_REFERENCE.md are your friends :)

Looks good to me - may need tweaking when you start looking at responsive ads.

bjfletcher:
@andygnewman I've moved styling to ads and under the loaded selector :+1:

PR #1133, opened 2016-06-09
include flags in the section data

bjfletcher:
which will eventually go to the `myFT` component which needs the flags to determine whether to show the follow button

essentially the same fix as this:
https://github.com/Financial-Times/next-front-page/commit/4585e07792db2ba26f1888637eeb5513b4616b2d

/cc @leggsimon


andygnewman:
👍

PR #1135, opened 2016-06-09
Show lead opinion image

bjfletcher:
On the opinion page, for the lead story, we want to show the main image.


andygnewman:
👍  Not necessarily the long term solution, but should do for now.

PR #1136, opened 2016-06-09
Ad slots mid: non-responsive - stream rh top, responsive - below most read (w/ top section, else list pos 7)

andygnewman:
This trello card https://trello.com/c/8c9VX7bL

##Mid Ad Slot
- no ads below L breakpoint
- separately positioned ad slots for responsive / non-responsive ads

###Non-responsive
- right hand of stream page
- MPU / double MPU

###Responsive
- when top section is present, positioned below Most Read 
- else positioned after item 6 in stream list


views/article-list.html
```diff
@@ -101,8 +101,27 @@
...
-				Placholder for non-responsive DMPU ad slot
+			<div
+				class="stream__right-hand {{#if hasStreamListTitle}} stream__right-hand--has-title{{/if}}"
+				data-o-grid-colspan="12 L4 XL3"
+				{{#if hasStreamListTitle}} data-has-stream-list-title{{/if}}
+				>
+				<div
+					class="o-ads sidebar-advert"
+					data-o-ads-name="halfpage-desktop"
+					data-o-ads-targeting="pos=hlfmpu;"
```
bjfletcher: I don't see `hlfmpu` used anywhere else, so I'm guessing this is a new one? How about a more readable name, e.g., `halfway`?
Then it'll be consistent with `halfpage-desktop` for `name`. Just a thought.

bjfletcher:
LGTM :+1:

shared/components/responsive-ad/index.js
```diff
@@ -0,0 +1,27 @@
...
+				data-o-ads-targeting="pos=mid;"
+				data-o-ads-center="true"
+				data-o-ads-label="false"
+				data-o-ads-formats-default="false"
+				data-o-ads-formats-small="false"
+				data-o-ads-formats-medium="false"
+				data-o-ads-formats-large="Responsive"
+				data-o-ads-formats-extra="Responsive"
+				data-o-ads-collapse-empty="true"
+				data-o-ads-lazy-load="false"
```
adgad: I think you can still lazy load this - o-ads will disable the lazyloading automatically if the top ad is part of a responsive master/companion.

PR #1139, opened 2016-06-10
Section Ads separate out ad slots for S vs M in O&A

andygnewman:
@bjfletcher 

Was reviewing overall ad-placement on streams.
The unit after Opinion & Analysis should also accept Responsive units, so have added that in.

Also o-ads can only cope with two ad position names on a page (top and mid).

On Default and S screen sizes, this ad unit effectively becomes the top pos (as no masthead shown), and a further ad appears lower down (trello card 137).

So I have split this unit into two components, one tied to Default & S sizes (where pos = top), the other for Medium size (where pos=mid).


bjfletcher:
Nice :+1:

PR #1141, opened 2016-06-10
simplify layout, removing some rubbish & now looking much better on IE9 -> IE11

bjfletcher:
shame GitHub doesn't show diff well when much of the line changes were actually indention changes

essentially what I've done:

* remove `<div class="stream{{#if brand}} stream--{{brand}}{{/if}}">` as it's no longer used
* add a `o-grid-row` where it should have been there (previously there were col -> col -> row, now col -> row -> col -> row)
* remove trending topics padding hack


views/article-list.html
```diff
@@ -34,127 +34,127 @@
 {{/if}}
 
 <div role="main" id="site-content" class="o-grid-container" data-concept-id="{{conceptId}}">
-	<div class="stream{{#if brand}} stream--{{brand}}{{/if}}">
```
andygnewman: Are you sure this isn't used any more? Can you check for any styling for brexit brand for example?

bjfletcher: https://github.com/search?q=org%3AFinancial-Times+stream--&type=Code

seems to suggest no.

Also... https://github.com/Financial-Times/next-stream-page/blob/master/shared/components/stream/_main.scss#L1

andygnewman: OK

ironsidevsquincy:
Ignore whitespace changes - https://github.com/Financial-Times/next-stream-page/pull/1141/files?w=1

bjfletcher:
@ironsidevsquincy really really happy to have that capability, thanks for telling me about it!

PR #1144, opened 2016-06-13
Get told by Sentry and not the editorial team, of issues

bjfletcher:
We want to be notified before editorial team come knocking on our door with problems.

This PR should ensure that Bug Duty gets notified immediately of any errors with the stream page app.


i-like-robots:
LGTM 👍

server/controllers/stream.js
```diff
@@ -95,6 +95,10 @@ module.exports = (req, res) => {
 	// Or fail gracefully
 	.catch(err => {
 		logger.error(err);
-		res.sendStatus(err.status || 500);
+		if (err.status === 404) {
```
matthew-andrews: neat if this works, where does this get set? :)

matthew-andrews:
also LGTM - but would like to understand a bit more how it works

bjfletcher:
This :) https://github.com/Financial-Times/next-stream-page/blob/master/server/controllers/stream.js#L59

PR #1146, opened 2016-06-15
remove ad from related-content component as it's now inserted in stream

bjfletcher:
I hope that is what Andy would have done if he was here.

FYI @andygnewman https://trello.com/c/heEAnkuC/190-sections-streams-too-many-ads


andygnewman:
Yes was exactly what I would have done (see comment at bottom of this card -  https://trello.com/c/EjdiNowq/167-video-related-content-on-stream-section-pages)

PR #1148, opened 2016-06-16
Section pages performance improvements - above-the-fold section head.css & lazy-loading images

bjfletcher:
* use lazy loading for images in all sections
* include only above-the-fold CSS in critical CSS for section pages
* note that when we finish stream works, we will either
a) move above-the-css CSS stream stuff into critical; or
b) have separate head.css's for section and stream pages

head.css is now 194k, with potential for a further reduction to around 145k when https://github.com/Financial-Times/n-ui/issues/125 is resolved


client/main.scss
```diff
@@ -4,7 +4,7 @@
 	preset: complete,
 	critical: true,
 	buttons: true,
-	myft: 'critical'
+	myft: 'critical' // at least until myFT splits up and we can get just the buttons
```
bjfletcher: I see that @wheresrhys doesn't single-quote the string values, whereas @ironsidevsquincy does. Which approach should we use?

wheresrhys: single quotes is clearer. I'm lazy

ironsidevsquincy: Strings in sass really freak me out, they're rather confusing. Sometimes unquoted strings can get converted to other things, e.g. `blue` (unquoted) becomes the colour blue, so to be safe I always quote

client/main.scss
```diff
@@ -14,20 +14,22 @@
...
+@import 'components/sub-header/main';
+@import 'n-section/main';
+@import 'components/section/main';
+
+@include nUiCriticalEnd;
+
 $o-icons-image-service-base-url: 'https://next-geebee.ft.com/image';
 @import 'o-share/main';
 $o-expander-is-silent: false;
 @import 'n-ui/expander/main';
```
ironsidevsquincy: can add `expander: true` into the above configuration list

config/layouts/lead-story.js
```diff
@@ -15,7 +15,8 @@ export default [{
 				show: { default: true, M: false, L: true },
 				colspan: { default: 12 }
 			},
-			image: createImageConfig({ default: 12, L: 9 }, { default: 'top', M: 'right' }, [280, 400, 517, 659])
+			image: createImageConfig({ default: 12, L: 9 }, { default: 'top', M: 'right' }, [280, 400, 517, 659]),
+			lazyLoadImages: false
```
ironsidevsquincy: Is this the first story on the page? Might be worth not lazy-loading this one, as it's the most prominent (and in case loading fails for whatever reason)

bjfletcher: I agree, I think if we set it to `false`, it won't be lazy-loaded, right?

client/main.scss
bjfletcher: can I clarify, remove these:

```
$o-expander-is-silent: false;
 @import 'n-ui/expander/main';
```

and have `expander: 'critical'` in the n-ui configuration, is that what you meant?

config/layouts/lead-story.js
ironsidevsquincy: Oh, sorry, didn't see it was `false`. Yeah, or just leave it off completely

client/main.scss
ironsidevsquincy: Oh yes, you're right, should be `critical`

PR #1149, opened 2016-06-16
Expander feature is critical

ironsidevsquincy:
For the 'Related to your search' dropdown on the search results

Another example of a feature, added to the CSS, that's only used in a single place...


bjfletcher:
@wheresrhys that's used only in stream pages, so how about we put that under a comment "for stream"

down the road, we can split section & stream up

PR #1150, opened 2016-06-16
Tidy up ad code

adgad:
/cc @bjfletcher @andygnewman 
*Remove all the old unused layouts
* Move some of the common classes to n-ui
* Flag to support reserving billboard heights (instead of leaderboard heights)


client/components/ads/_main.scss
```diff
@@ -26,29 +14,14 @@ $ad-medium-breakpoint: 760px;
...
 .above-header-advert {
+	display: none;
 	@media (min-width:$ad-medium-breakpoint) {
 		display: block;
-		min-height: 90px;
-		height: 90px;
 		padding: 10px 0;
 		line-height: 0;
-		white-space: pre;
 		text-align: center;
```
bjfletcher: I wonder if we want these `display`, `line-height` and `text-align` properties?

bjfletcher:
If you're happy & confident with these changes, then go ahead. Otherwise, wait for Monday when @andygnewman can have a look at this too? He's been doing a lot of work around ads here.

client/components/ads/_main.scss
adgad: We need the display, but I shall remove the others.

PR #1152, opened 2016-06-17
new stream item card

bjfletcher:
The cards themselves now looking the same as Sari's design.  See screengrab.  This uses lazy-loading technique for a snappier page load and scroll performance.

image

Some more work to do - remove unnecessary rules, align with More News headline, style the date on the left, etc.


shared/components/card-new/index.js
```diff
@@ -1,75 +1,40 @@
...
+			id: item.id,
+			title: item.title,
+			published: item.published,
+			lastPublished: item.lastPublished,
+			standfirst: {
+				show: { default: true },
+				text: item.summary
+			},
+			tag: item.primaryTag,
+			image: item.primaryImage ? {
```
andygnewman: Wondering why we're recreating an article model here as well as https://github.com/Financial-Times/next-stream-page/blob/master/server/models/article.js

Could we have all this in the one place (plus it already had the right widths and sizes).

ironsidevsquincy:
Bit worried that we now have a different card design than the front page. Can we not have the same style of left-aligned cards everywhere?

andygnewman:
@ironsidevsquincy agree that it would be better to have a single design.

However can understand why there might be some differences as we're using the standfirst (not on home page) and we need headline above image to allow for consistent vertical scanning when image not present.

bjfletcher:
@ironsidevsquincy in addition to what Andy said, it looks cheap, old & ugly when you've got images stacked up on the left with wasted space in between, and lots of text on the right to include the four different texts - tag, headline, standfirst, and date.

ironsidevsquincy:
I agree, but all these issues are applicable on the front page, so if this is a better solution we should apply this everywhere

bjfletcher:
@ironsidevsquincy sounds good to me :) to use it, just change `position` to use `embedded` like so:

```position: { default: 'embedded' }```

in the model. FYI, @onishiweb mentioned that myFT is also working on a replacement card style and he was going to help collate all the new and replacement variants into `o-card`.

**Comment by ironsidevsquincy** [comment]
Could we not just update the left (and right?) positioned images to this new style, so it will roll out everywhere?

**Comment by bjfletcher** [comment]
image
image

Can I double check, you're thinking maybe it's better that we replace the first with the second everywhere?

**Comment by ironsidevsquincy** [comment]
I'm just keen on having one solution everywhere. Because of the issues you raised above, it sounds like the second card style is better than the first, so that's the one we should go with

**Comment by bjfletcher** [comment]
@ironsidevsquincy 

1. what'd we do with image that's on right?
2. I see some layouts, it is more consistent to have the bigger image (this new design has a smaller image)

I think we should keep both options, but that's my personal opinion... how should we get this decided? Ask Seri? (She designed it all.)

**Comment by ironsidevsquincy** [comment]
Yes, think we need to stress to design that consistency = simpler to build and maintain. And agreed, there might be a case for having both

File: server/models/article.js [hunk 2]
```diff
@@ -15,24 +9,32 @@ module.exports = opts => {
...
+					template: 'card-new',
+					standfirst: {
+						show: { default: true },
+						text: model.summary
+					},
+					tag: model.primaryTag,
+					image: model.primaryImage ? {
+						url: model.primaryImage.url,
+						position: { default: 'embedded' },
+						widths: [200, 500],
```
andygnewman: Was there a reason for changing the widths and sizes from the original code? Those had been crafted to the list layout?

andygnewman:
Looking pretty neat. Just the one comment about the change to image widths / size.

server/models/article.js
bjfletcher: @andygnewman original code put back in, not sure why I took it out having thought about it.

PR #1154, opened 2016-06-20
fix chevron issue with the related links in lead story

bjfletcher:
@ironsidevsquincy kindly investigated and identified that this was because we included `o-card` in `main.scss`

So I've just tried that, and indeed that fixed it. Genius. :)


andygnewman:
LGTM

PR #1155, opened 2016-06-20
[WIP] Stream Properties

andygnewman:
**Objective:**
Store stream properties in an object that can be used to feed spoor data.
Example properties would be along the lines of
```
{
    isPage: 2,
    hasTopStories: true,
    hasPhotoDiary
}
```

These properties would be run as a transform (the last one being run).
These properties could be used for if statements in mustache templates or other rendering branching as well as for populating the data.
Would refactor out the setting and usage of such properties across the app.


server/transforms/stream-properties.js
```diff
@@ -0,0 +1,44 @@
...
+			streamProperties.isLive = templateData.hasOwnProperty('isLive');
+			streamProperties.hasLeadStory = templateData.hasOwnProperty('leadStory');
+			streamProperties.hasTopStories = templateData.hasOwnProperty('topStories');
+			streamProperties.hasOpinionAndAnalysis = templateData.hasOwnProperty('opinionAndAnalysis');
+			streamProperties.hasMostRead = templateData.hasOwnProperty('mostRead');
+			streamProperties.hasOpenInStream = templateData.hasOwnProperty('openInStream');
+			streamProperties.hasTrendingTopics = templateData.hasOwnProperty('trendingTopics');
+			streamProperties.hasBranding = templateData.hasOwnProperty('branding');
+			streamProperties.hasMarketsData = templateData.hasOwnProperty('marketsData');
+			streamProperties.isSpecialReport = templateData.hasOwnProperty('isSpecialReport');
```
bjfletcher: I'm wondering if we should put all that to the individual modules, for example, in Top Stories transform:

`templateData.metadata.topStories = true;`

if you prefer to keep this centralisation approach, cleaner code may be:

```
const streamProperties = {
  isLive: !!templateData.isLive,
  hasLeadStory: !!template.leadStory
  ...
}
```

andygnewman: We could delegate to each module, however I'm wondering whether every module actually is in control of whether it is presented or not? Eg. we may do some resolutions elsewhere that discard some modules from the presentation.

So actually instead of it being a transform, it may be better to call it from render-stream, so we're totally clear on what elements will appear on the page (subject to any client side ones).

bjfletcher: @andygnewman aha, good point... how about `templateData.leadStory.meta = { isPresent: true }` and then for spoor, just `const meta = templateData.map(t => t.meta).flatten()`?

on your 2nd paragraph, do you mean for the multiple CSS files? The approach DH is currently using in his PR?

PR #1156, opened 2016-06-20
add middleware to detect if permalink parameter has been passed. if s…

VladDubrovskis:
…o, get url and redirect
/cc @andygnewman @bjfletcher 

Let me know your thoughts on this one, now that I have done it not too sure if middleware was best place for it, but theoretically it is quite hight up on request chain to redirect before hitting the rest of the application?


test/server/middleware/blogs-premalink.test.js
```diff
@@ -0,0 +1,63 @@
...
+					p: '521553'
+				}
+			};
+
+
+			return createInstance(params);
+		});
+
+		it('returns a redirect header response', (done) => {
+			setTimeout(() => {
```
andygnewman: Not a big fan of timeouts in tests. We've not used them for any other tests in streams where we're getting promises / api responses. Example: https://github.com/Financial-Times/next-stream-page/blob/master/test/server/lib/fetch-api-v2.test.js

test/server/middleware/blogs-premalink.test.js
```diff
@@ -0,0 +1,63 @@
...
+		nock.cleanAll();
+	});
+
+	context('success', () => {
+
+		beforeEach(() => {
+
+			nock('http://www.ft.com')
+				.get('/fastft/api/get_post/?id=521553&api_key=1234')
+				.reply(200, fixtureBlogsApiResult);
```
andygnewman: If you're not asserting anything about the fixtureBlogsApiResult, you could simplify it to something such as {post: {url: example.com}}

server/middleware/blogs-permalink.js
```diff
@@ -0,0 +1,18 @@
+const fetchres = require('fetchres');
+module.exports = (req, res, next) => {
+	if(req.query.p) {
```
andygnewman: is this specific enough? if a p query string string in on a non-fastFT stream for any reason it would do the wrong thing. Not sure if other blogs (eg. alphaville have a similar concept?)

test/server/middleware/blogs-premalink.test.js
VladDubrovskis: Fair enough, not a big fan of timeouts either, just wanted to get feedback on this. Just was not sure about the Promise being returned from middleware

server/middleware/blogs-permalink.js
VladDubrovskis: is there any helper method that can tell it is FastFT? or do i need to check for the stream id?

andygnewman: https://github.com/Financial-Times/next-stream-page/blob/master/server/config/profiles.js#L41

server/middleware/blogs-permalink.js
```diff
@@ -0,0 +1,31 @@
+const fetchres = require('fetchres');
+
+const isFastFT = url => url.indexOf('NTlhNzEyMzMtZjBjZi00Y2U1LTg0ODUtZWVjNmEyYmU1NzQ2-QnJhbmRz') > -1;
```
VladDubrovskis: passing this a "req.url" as I think the middleware will not have the "thingId" as it would not have hit the express routing yet

server/middleware/blogs-permalink.js
```diff
@@ -0,0 +1,31 @@
...
+
+module.exports = (req, res, next) => {
+	return new Promise((resolve, reject) => {
+		if(req.query.p) {
+			return fetch(`http://www.ft.com/fastft/api/get_post/?id=${req.query.p}&api_key=${process.env.BLOGS_API_KEY}`, {timeout: 500})
+				.then(fetchres.json)
+				.then((response) => {
+					if(response && response.status && response.status === 'ok' && response.post && response.post.url && isFastFT(req.url)) {
```
andygnewman: Would it be better to check isFastFT in the original if
`if(req.query.p && isFastFT(req.url))`

VladDubrovskis: yeah, makes sense, so we don't do unnecessary fetch

andygnewman: fetchres already checks for response, status etc https://github.com/matthew-andrews/fetchres/blob/master/src/fetchres.js#L3

VladDubrovskis: this is just checking the actual response from FastFT api, just making sure structure is there and just a little defensive coding

server/middleware/blogs-permalink.js
```diff
@@ -0,0 +1,31 @@
+const fetchres = require('fetchres');
+
+const isFastFT = url => url.indexOf('NTlhNzEyMzMtZjBjZi00Y2U1LTg0ODUtZWVjNmEyYmU1NzQ2-QnJhbmRz') > -1;
+
+module.exports = (req, res, next) => {
+	return new Promise((resolve, reject) => {
```
andygnewman: Does it all need to be wrapped in a promise in this way?

VladDubrovskis: yeah, otherwise fetch returns a promise, and other cases do not. To keep it consistent in terms of what it returns: Always a promise. Also allows to remove the setTimeout from tests

test/server/middleware/blogs-premalink.test.js
```diff
@@ -0,0 +1,92 @@
...
+
+		it('returns a redirect header response when parameter passed to FastFT stream', (done) => {
+			let params = {
+				url: '/stream/brandId/NTlhNzEyMzMtZjBjZi00Y2U1LTg0ODUtZWVjNmEyYmU1NzQ2-QnJhbmRz',
+				query: {
+					p: '521553'
+				}
+			};
+
+			createInstance(params).then(() => {
```
andygnewman: You can use this type of pattern in the tests when a promise is returned. https://github.com/Financial-Times/next-stream-page/blob/master/test/server/lib/concordance.test.js#L55 so you won't need the done (or timeouts)

bjfletcher:
How about this?

```
const isFastFT = url => url.indexOf('NTlhNzEyMzMtZjBjZi00Y2U1LTg0ODUtZWVjNmEyYmU1NzQ2-QnJhbmRz') !== -1;

module.exports = (req, res, next) => {
	if (req.query.p && isFastFT(req.url)) {
		const url = `http://www.ft.com/fastft/api/get_post/?id=${req.query.p}&api_key=${process.env.BLOGS_API_KEY}`;
		return fetch(url, { timeout: 500 })
			.then(fetchres.json)
			.then(response => {
				if (response && response.status && response.status === 'ok' && response.post && response.post.url) {
					res.redirect(302, response.post.url);
				} else {
					throw new Error('wassup, here?');
				}
			})
	}
	next();
}
```

bjfletcher:
A few thoughts... once we've decided to redirect, probably best to not call `next()`

VladDubrovskis:
soo, something like this:
```
const fetchres = require('fetchres');
const isFastFT = url => url.indexOf('NTlhNzEyMzMtZjBjZi00Y2U1LTg0ODUtZWVjNmEyYmU1NzQ2-QnJhbmRz') !== -1;

module.exports = (req, res, next) => {
	if (req.query.p && isFastFT(req.url)) {
		const url = `http://www.ft.com/fastft/api/get_post/?id=${req.query.p}&api_key=${process.env.BLOGS_API_KEY}`;
		return fetch(url, {
				timeout: 500
			})
			.then(fetchres.json)
			.then(response => {
				if (response && response.status && response.status === 'ok' && response.post && response.post.url) {
					res.redirect(302, response.post.url);
					res.end(); //think this is required to stop falling futher down the execution chain?
				} else {
					next();
				}
			})
	} else {
		next();
	}
}
```

VladDubrovskis:
though then we get inconsistency about sometimes we return a promise and sometimes we just fall through. First time writing middleware so if this is fine, more than happy to keep it that way :)

server/middleware/blogs-permalink.js
```diff
@@ -0,0 +1,23 @@
...
+	if (req.query.p && isFastFT(req.url)) {
+		const url = `http://www.ft.com/fastft/api/get_post/?id=${req.query.p}&api_key=${process.env.BLOGS_API_KEY}`;
+		return fetch(url, {
+				timeout: 500
+			})
+			.then(fetchres.json)
+			.then(data => {
+				if (data && data.status && data.status === 'ok' && data.post && data.post.url) {
+					res.redirect(302, data.post.url);
+					res.end();
```
bjfletcher: > res.redirect() indeed calls res.end() itself

http://stackoverflow.com/a/15572102/722180

server/middleware/blogs-permalink.js
```diff
@@ -0,0 +1,23 @@
...
+				} else {
+					next();
+				}
+			})
+			.catch(next)
```
bjfletcher: why running `next()` when there's an error?

VladDubrovskis: so if there fetch request messes up it falls through to next thing in the chain? Or leaving this out will have same effect?

bjfletcher: this error happens only within the `req.query.p && isFastFT(req.url)` domain, and so, yes, I think we should allow the error to bubble up to Sentry, for you to investigate... rather than the error getting swallowed up and keeping quiet about it

VladDubrovskis: aah, I see, that makes sense. I guess my thinking was one of defensive coding, to anticipate things going wrong and add safeguards in advance, but missed to notice the point you making .
Will remove the catch

bjfletcher:
LRGTM :+1: 

(Just one last little bit thing... add some more explanation to `HACKAROUND FastFT permalinks working on Next - starts here` as to when we should remove this hack, then it'll be super good to go! :))

andygnewman:
I think I can give it a thumbs up now! Good work @VladDubrovskis

andygnewman:
👍

PR #1160, opened 2016-06-21
Photo diary reposition and redesign

onishiweb:
From: https://trello.com/c/eSOcKu3f

Moves the photo diary section out of the streams list and to the bottom of the World stream page. Design has also been updated:

screen shot 2016-06-21 at 14 05 02


andygnewman:
Looking good. Nice clean up of the redundant code. Might be an idea to run through it with Simon sat next to you on local to see if he spots any design issues, but otherwise 👍

onishiweb:
Will do. Just need to run through the browser tests for it quickly and make sure they're ok.

Also added one last test for the removal of photo diary items in the new photo diary module if you could take a look at that @andygnewman 

Cheers

onishiweb:
Checked in with Simon and browser tested as well. All looking good.

test/server/lib/photo-diary.test.js
```diff
@@ -1,21 +1,43 @@
...
 	});
 
-	it('should return 4 photo diary elements', function() {
-		result[24].items.length.should.eq(4);
+	it('should not return if less than 4 photo diary elements', () => {
+		should.not.exist(resultShort);
+	});
+
+	it('removes photo diary entries from stream items', () => {
+		const photoDiaryItemsTotal = streamFixture.filter(element => {
```
andygnewman: you could just use `result.length` instead of `photoDiaryItemsTotal`

onishiweb: I don't think I can, as I need to get all the photo diary elements in the stream, whereas `getPhotoDiaries` will only return `.length == 4` or `undefined` if there's <4. Had to run the filter (and modify it slightly) again, to get the true number that would be removed from the stream.

andygnewman: Yep, good point, the items in the gallery and the removed items are not necessarily equal.

bjfletcher:
Really nice code. :+1:

Worth letting frontpage people (@ironsidevsquincy @andygout et al) know that we've got this, in case they'd like the same - in which case, we'd probably want to extract some of this code to the `n-section` (soon to be `o-section` - where we can utilise various components in a configurable way). A good example is the video component there.

https://github.com/Financial-Times/next-front-page/search?utf8=%E2%9C%93&q=video
https://github.com/Financial-Times/n-section/search?utf8=%E2%9C%93&q=video

onishiweb:
Yep sounds good, cheers Ben. I'm looking into the video stories next so will take a look at that too.

bjfletcher:
@onishiweb Cool, frontpage's video section is at the bottom of next.ft.com. Maybe once you've got this in our stream pages, have a think about whether you'd like to try and extract some of the photo work to a similar setup.

PR #1165, opened 2016-06-22
Explicitly turn critical css off in IE8

ironsidevsquincy:


bjfletcher:
Thanks @ironsidevsquincy :-) :+1:

PR #1167, opened 2016-06-22
New chronological date in stream list

bjfletcher:
image

image

image

image

image

* don't use new stream list on FastFT & News Feed (will be separate work)
* show just "Today", "Yesterday" or nothing for the relative time, with full date underneath
* styling as per design for all breakpoints


shared/components/chronological-date/index.js
```diff
@@ -0,0 +1,43 @@
+const React = require('react');
+const dateFormat = require('dateformat');
+
+function isToday (date) {
```
andygnewman: See you've created a new component rather than add to the existing o-date (which we'll need to retain anyway for the card) wondered whether they should be one component with variants or 2 separate components?

andygnewman: Also I'm kind of wondering whether the logic to create the relative label should sit in the lib file that creates summary date, so we can more easily test it and have the logic outside the presentation.

Then it would be simpler to have a single date presentation component?

bjfletcher: @andygnewman What would be really great to have is:

* a more dynamic version of this, where "Today" becomes "Yesterday" on the already-loaded webpage when the clock goes past midnight

* the chronological date to become sticky as you scroll down, so you'll always see the current date as you scroll down, if that makes sense? Max: "I tell you what would be really nice… is if the date stuck to the top of the page as you scrolled."

I see this - what we have in this PR - as a more temporary/interim MVP solution. What do you think we should do?

bjfletcher: Alice is looking at how we can do this using Origami's (not our) `o-date`. I think it makes sense to merge this and attempt to use Origami's `o-date` for a dynamic behaviour in a separate branch.

PR #1169, opened 2016-06-23
Add multiple CSS files

ironsidevsquincy:
There are now four different stylesheets for the four different type of stream page

 * stream (e.g. https://next.ft.com/world)
 * section (e.g. https://next.ft.com/global-economy)
 * fast-ft (only https://next.ft.com/fastft)
 * live (only https://next.ft.com/news-feed)

Deciding which stylesheet to use is based on https://github.com/Financial-Times/next-stream-page/pull/1169/files#diff-fdf8b96d1ea9b4c8a0e3795cb2784356R47


andygnewman:
I like the idea.

I've also been doing this PR https://github.com/Financial-Times/next-stream-page/pull/1155 which is primarily for feeding spoor with the customisation of the stream page.

This could be used to be more granular in loading in the required CSS for each of the elements on the page as it knows what should be there.

Just a thought.

bjfletcher:
Wooohoooooo. All my wishes have been happening today... :) Nice one!

server/lib/render-stream.js
```diff
@@ -42,6 +42,7 @@ module.exports = function (req, res) {
 			// show new top section if flag and all the data it needs is present
 			if (res.locals.flags.streamNewPages && ['leadStory', 'topStories', 'opinionAndAnalysis'].every(id => templateData.hasOwnProperty(id))) {
 				templateData.streamNewPages = true;
+				templateData.cssVariant = 'stream';
```
bjfletcher: is this meant to be section?

ironsidevsquincy: Yes, quite right. This should be in an `else`

ironsidevsquincy:
@andygnewman Interesting. How do you see that configuration object feeding into sass?

server/lib/render-stream.js
```diff
@@ -42,6 +42,12 @@ module.exports = function (req, res) {
 			// show new top section if flag and all the data it needs is present
 			if (res.locals.flags.streamNewPages && ['leadStory', 'topStories', 'opinionAndAnalysis'].every(id => templateData.hasOwnProperty(id))) {
 				templateData.streamNewPages = true;
+			} else if (templateData.isFastFt) {
+				templateData.cssVariant = 'fast-ft';
+			} else if (templateData.isAlive) {
+				templateData.cssVariant = 'live';
+			} else {
+				templateData.cssVariant = 'stream';
```
bjfletcher: ... and for section, look for whether templateData.leadStory exists

server/lib/render-stream.js
```diff
@@ -69,7 +75,8 @@ module.exports = function (req, res) {
 			if (templateData.photoDiaryItems) {
 				templateData.PhotoDiary = PhotoDiary;
 			}
-
+			console.log('############### IS ALIVE ###############');
+			console.log(templateData.isAlive)
```
bjfletcher: :boom: these two lines :)

server/transforms/live-stream.js
```diff
@@ -15,6 +15,8 @@ module.exports = () => {
 			res.templateData.isAlive = res.templateData.isLive
 				? true
 				: res.templateData.isAlive;
+
+			res.templateData.isFastFt = req.params.thingId === 'NTlhNzEyMzMtZjBjZi00Y2U1LTg0ODUtZWVjNmEyYmU1NzQ2-QnJhbmRz';
```
bjfletcher: or maybe `res.templateData.meta.isFastFt = ...` with `meta` in the middle, depending on how discussion in Andy's PR goes, I've never been a fan of centralised configuration and personally I prefer to have modular configuration like this

but that's my personal preference, it may be that we'll go with Andy's approach

either way, this PR looks good to me. :) 👍

server/lib/render-stream.js
ironsidevsquincy: keeping you on your toes...

ironsidevsquincy:
Ok, think this is ready to go. @bjfletcher and @andygnewman, could you take a look that https://github.com/Financial-Times/next-stream-page/pull/1169/files#diff-fdf8b96d1ea9b4c8a0e3795cb2784356R47 makes sense, and also I'm not missing any obvious `import`s from the sass files

server/lib/render-stream.js
andygnewman: I think that is covered in line 45.

server/transforms/live-stream.js
```diff
@@ -4,6 +4,7 @@ module.exports = () => {
 	return class {
 		pre(req) {
 			this.isFirstPage = req.perPage.page === 1;
+			this.isFastFt = req.params.thingId === 'NTlhNzEyMzMtZjBjZi00Y2U1LTg0ODUtZWVjNmEyYmU1NzQ2-QnJhbmRz';
```
andygnewman: The location of this feels pretty arbitrary and perhaps a little conflicted with the purpose of the transform.

Wondering whether it can go into the controller or be part of renderStream?

ironsidevsquincy: Isn't it in keeping with what else is happening in this file (setting a flag if it's the first page/if it's 'alive')?

andygnewman: The transform is about setting the live nature of the stream, which could apply to any stream (e.g. also used by News Feed and could be expanded to others in future), not explicitly FastFT,  plus there are 3 other transforms that apply to FastFT, so why choose this one?

If the CSS config is related purely to the live nature then I would rename it, but if it's because of the combined attributes of FastFT I would recommend putting it elsewhere.

PR #1173, opened 2016-06-24
Have a more descriptive error for when there is no content to make debugging easier

matthew-andrews:


commuterjoy:
+1

bjfletcher:
+1

matthew-andrews:
i don't understand the error, i don't think it's my PR's fault…  it also fails on master 😞

PR #1174, opened 2016-06-24
Move to use list API for curated content with stream and list type

andygnewman:
Trello: https://trello.com/c/GolJbppS

Objective: Give Editorial easier control over which section pages have a curated Top Section.

Key Features:
- [x] Enable Top Stories to get list of stories from either Pages API (for existing Preditor maintained Section Pages) or from List API using Stream ID and List Type.
- [x] Migrate curated Opinion and Analysis to get list of stories from List API using Stream ID and List Type (rather than hardcoded list of List IDs mapped to stream IDs)

Other Features Implemented in this PR:
- [x] Balance split between first and second column of top stories (when no O&A) - ie. if there were 7 top stories, there would be 1 in lead, 3 in column one, 3 in column 2 (rather than 1, 5, 1 in current).
- [x] Be clear in the data when Opinion and Analysis exists in the top section (2nd Column was previously always called Opinion and Analysis regardless of content) - this is vital to the spoor work.
- [x] Refactoring of the code to split out the getting of the data from the assembly of the data.


server/lib/opinion-and-analysis.js
```diff
@@ -0,0 +1,38 @@
...
+
+  const commentTags = ['Comment', 'Special Report', 'Analysis'];
+
+  return streamItems.filter(item => item.metadata && itemGenreIs(item, commentTags))
+    // filter out items we already have in top stories
+    .filter(item => summaryIds.indexOf(item.id) === -1)
+    .slice(0, 4)
+    .map(item => articleModel(item, 'article-card'));
+
+};
```
bjfletcher: nice work there - very tidy & readable

maybe use ES6 module syntax...

```
export fromApi = () ....
export fromStreamList = () ...
```

PR #1175, opened 2016-06-27
Fix stickies test

andygnewman:
@i-like-robots @bjfletcher @matthew-andrews @callumlocke 

This fixes the stickies test that's failing master on stream app.

Thanks to @i-like-robots for an example of mocking a Poller in n-navigation.


test/stubs/poller-stub.js
```diff
@@ -0,0 +1,13 @@
+'use strict';
+
+const Poller = require('ft-poller');
+const sinon = require('sinon');
+const stickiesManifest = require('../fixtures/stickies-manifest');
+
+let instance = sinon.createStubInstance(Poller);
+
+module.exports.Poller = function (opts) {
+  instance.start = () => Promise.resolve(opts.initialRequest);
```
i-like-robots: lint

matthew-andrews:
+1 nice

test/stubs/poller-stub.js
andygnewman: ta

PR #1177, opened 2016-06-27
add .scss-lint.yml

bjfletcher:
for Sass linting within IDE/editor

/cc @ironsidevsquincy @andygnewman


ironsidevsquincy:
No need to check it in - just put it in the `.gitignore` and `make install` will download it for you automatically - https://github.com/Financial-Times/next-front-page/blob/master/.gitignore#L28

bjfletcher:
That .gitignore trick is confusing and way too hacky. I think it makes better sense to have some `install.yml` with a list of files we want to download and install, with a big fat comment at the top to remind people to mirror content in .gitignore.

ironsidevsquincy:
Hmmm, possibly. Though this takes care of all that for you automatically, i.e. you don't have to remember to put them in the ignore, as they have to be in there to download them

You could open an issue in [n-makefile](https://github.com/Financial-Times/n-makefile), get Matt A's thoughts (he wrote it)

PR #1178, opened 2016-06-29
update way we init the videos on stream page and to enable running ad…

VladDubrovskis:
…s on oVideo.

Should call the init method which will make a call to factory. Required by oVideo in order to run advertising

/cc @andygnewman @bjfletcher


AlbertoElias:
👍

andygnewman:
OK by me, I think, I suspect my knowledge of o/nVideo is probably inferior to that of @VladDubrovskis & @AlbertoElias

PR #1179, opened 2016-06-29
Update to use o-date's dynamic relative time

bjfletcher:
The `(Yesterday|Today|)` is now done in `o-date`: https://github.com/Financial-Times/o-date/pull/56

PR #1180, opened 2016-06-29
Add flag for save button

bjfletcher:
For the new stream list.

Please note that this branches off #1179


bjfletcher:
Green build: https://circleci.com/gh/Financial-Times/next-stream-page/3285

bower.json
```diff
@@ -3,20 +3,22 @@
...
+    "o-date": "^2.4.0",
     "o-fetch-jsonp": "^1.0.1",
     "o-icons": "^4.1.2",
     "o-share": "^2.1.3",
     "o-video": ">=0.1.0 <1"
   },
   "resolutions": {
-    "n-card": "^5.4.0",
+    "o-date": "^2.4.0",
+    "n-card": "^5.5.0",
```
ironsidevsquincy: why are these needed?

n-makefile.json
```diff
@@ -13,12 +13,14 @@
...
-      "@financial-times/n-section",
       "@financial-times/n-image",
+      "@financial-times/n-section",
       "@financial-times/n-ui",
+      "n-card",
       "n-image",
       "n-section",
-      "n-ui"
+      "n-ui",
+      "o-date"
```
ironsidevsquincy: shouldn't need any of these

bjfletcher: `o-date` API is used for formatting the date server-side before it goes to client side (so you don't see any change as `o-date` responsive stuff loads)

bower.json
bjfletcher: no longer needed - removed now

n-makefile.json
ironsidevsquincy: Hmmm, we already do that in n-card, and it isn't needed then - https://github.com/Financial-Times/n-card/blob/master/components/timestamp/timestamp.js

Also, like n-card, shouldn't that formatting happen client side? As could have any old time on the server (it fades in the time so it's less jarring)

PR #1181, opened 2016-06-29
reinstate oVideo styles

VladDubrovskis:
/cc @ironsidevsquincy 

we plan to release it to a subset of users at some point this/next week to gather feedback


client/main-live.scss
```diff
@@ -25,6 +25,7 @@ $output-critical: true !default;
 @include nUiCriticalEnd();
 
 @import 'n-video/main';
+@import 'o-video/main';
```
ironsidevsquincy: Actually, no need for any video styling in this file, as no videos on https://next.ft.com/news-feed

ironsidevsquincy: (which is the only place this stylesheet is used)

ironsidevsquincy:
👍

PR #1185, opened 2016-07-01
Put open in stream on photo diary

matthew-andrews:
screen shot 2016-07-01 at 18 26 04

20 minute hack, just for fun…  feel free to say no @mmockett.

If yes I'll need to tweak how one thing is done.


server/lib/render-stream.js
```diff
@@ -45,7 +45,7 @@ module.exports = function (req, res) {
 			if (res.locals.flags.streamNewPages && ['leadStory', 'topStories', 'opinionAndAnalysis'].every(id => templateData.hasOwnProperty(id))) {
 				templateData.streamNewPages = true;
 				templateData.cssVariant = 'section';
-			} else if (req.params.thingId === 'NTlhNzEyMzMtZjBjZi00Y2U1LTg0ODUtZWVjNmEyYmU1NzQ2-QnJhbmRz') {
+			} else if (req.params.thingId === 'NTlhNzEyMzMtZjBjZi00Y2U1LTg0ODUtZWVjNmEyYmU1NzQ2-QnJhbmRz' || req.params.thingId === 'NTQ2MGRhM2UtMGNlOC00NDBjLTgyNWEtY2VmMWZkMjk5NDdk-QnJhbmRz') {
 				templateData.cssVariant = 'fast-ft';
```
matthew-andrews: is there a reason why this `cssVariant` stuff can't go inside the `config/profile.js` file?

mmockett:
Nice! Why not...

matthew-andrews:
OK this is ready for @andygnewman / @bjfletcher to review.  I've also spotted [a bug that affect this fairly badly — but also affects Fast FT](https://github.com/Financial-Times/next-article/issues/1451) so I think that can be looked separately.

.gitignore
```diff
@@ -42,6 +44,10 @@ npm-debug.log
 /public/main-fast-ft.css.map
 /public/main-fast-ft-ie8.css
 /public/main-fast-ft-ie8.css.map
+/public/main-photo-diary.css
```
andygnewman: Would suggest we just use the fast ft css for this rather than create a new one.

matthew-andrews: it's open in stream and live stream but maybe that's ok?

matthew-andrews: @ironsidevsquincy / @wheresrhys do you have any advice about this?

ironsidevsquincy: It's a balance. We don't want to have 100s of ever-so-slightly varying stylesheets, as that will defeat our cache, but we also want to give relatively 'distinct' pages their own files. 

I'll leave it up to you to decide whether the slight redundancy in using the fast-ft one is worth it (sorry, wishy-washy answer)

matthew-andrews: ¯\_(ツ)_/¯ 

so rename `fast-ft` to `live-opening-stream`?

wheresrhys: Possibly having ahead.css file few different head.css's but one catch all main.css would be a good thing to aim for. Most (all?) the variance between the different pages is above the fold, so main.css would be fairly similar for each anyway

Not necessarily achievable at this juncture though

matthew-andrews:
`fast-ft` is now `live-opening` — is it ok? @andygnewman

andygnewman:
👍

PR #1189, opened 2016-07-04
Use coco endpoint for concordance

andygnewman:
capi end point not updated with concordance, so going through coco for this and stream branding.


matthew-andrews:
nice, well not nice, but +1

PR #1192, opened 2016-07-05
Add: Authors transform to provide data for subheader

andygout:
cc @andygnewman


server/transforms/authors.js
```diff
@@ -0,0 +1,39 @@
+const concordance = require('../lib/concordance');
+const fetchApiV2 = require('../lib/fetch-api-v2');
+
+const getAuthorData = authorTMEId =>
+	concordance.tmeToV2(authorTMEId)
+		.then(authorUuid => authorUuid ? fetchApiV2({ path: `/people/${encodeURIComponent(authorUuid)}` }) : null);
+
+const formatAuthorData = author => {
+	const authorObject = {
+		strapline: author.memberships[0].title,
```
andygout: This doesn't feel quite right and think a talk regarding the API structure will help clarify.

andygnewman:
Neat. Let's wait until we get clarity on the api before finishing it off.
May be worth marking it as [WIP] or [DO NOT MERGE] for the time being.

server/transforms/authors.js
```diff
@@ -0,0 +1,37 @@
...
+module.exports = () => {
+	return class {
+		pre (req) {
+			this.author = getAuthorData(req.params.thingId);
+		}
+
+		post (res) {
+			return Promise.resolve(this.author)
+				.then(author => {
+					if (author) res.templateData.subHeader = formatAuthorData(author);
```
andygout: Until the People API is ready this line will simply do nothing (`author` value is `null`), so seems like it is safe to deploy in the meantime; it will simply do nothing.

It's also possible that the strapline value is acquired from a different attribute value of the returned `author` but can always fix that when required.

server/transforms/authors.js
```diff
@@ -28,9 +27,14 @@ module.exports = () => {
 		}
 
 		post (res) {
+			const subHeaderTypeObj = { type: 'author' }
 			return Promise.resolve(this.author)
 				.then(author => {
-					if (author) res.templateData.subHeader = formatAuthorData(author);
+					if (author) {
```
andygnewman: How about;

```
        .then(author => {
            res.templateData.subheader = { type: 'author'};
            if (author) {
                Object.assign(res.templateData.subHeader, formatAuthorData(author));
            }
        });
```

andygnewman:
👍

PR #1194, opened 2016-07-05
Switch to using .url property for article URLs on cards

matthew-andrews:
dependent on https://github.com/Financial-Times/n-content-model/pull/6 and this returning 0 results: http://search-nextcontent-22gafrfc246u3xwlec4nvh7gre.us-east-1.es.amazonaws.com/v3_api_v2/item/_search?q=_missing_:url.

(tests have a better chance of passing when the above is done^^)


matthew-andrews:
@andygnewman does this look ok?

andygnewman:
LGTM 👍

PR #1196, opened 2016-07-05
fix IE9 card alignment issue & remove some old, unnecessary code

bjfletcher:
IE9 flattens the `div` which relies on `flexbox` for spacing. This solution seems good.


andygnewman:
👍

PR #1197, opened 2016-07-05
make stream trending topics layout consistent with section

andygnewman:
Trello Card: https://trello.com/c/WpZAdyq5


client/main-stream.scss
```diff
@@ -15,6 +15,7 @@ $output-critical: true !default;
 @include nUiCriticalStart();
 
 @import 'n-card/main';
+@import 'n-section/main';
```
bjfletcher: would it make better sense to import just the meta stylesheet?

PR #1199, opened 2016-07-05
Custom Fragments bring into line with new stream list design

andygnewman:
cc: @callumlocke 

Trello Card: https://trello.com/c/l3pAxYNB

Key Objectives:
- [x] If custom fragment indicated to be at index 0, ensure it sits outside the stream list and can go full width
- [x] if at any other index retain in stream list
- [x] add some tests for the custom fragment transform
- [x] ensure this works with both the old and new stream list (flag: newStreamList)


views/article-list.html
```diff
@@ -26,6 +26,9 @@
 {{/if}}
 
 <div role="main" id="site-content" class="o-grid-container" data-concept-id="{{conceptId}}">
+	{{#with customFragment}}
+		{{{reactRenderToString ../CustomFragment spread=@this}}}
+	{{/with}}
```
bjfletcher: maybe reduce the above three lines to:

```{{{reactRenderToString CustomFragment spread=@customFragment}}}```

no worries if that doesn't work

bjfletcher:
LGTM

views/article-list.html
andygnewman: Unfortunately doesn't work as I only add the CustomFragment component in when we know there is a customFragment object.

PR #1203, opened 2016-07-07
remove summary date in favour of user-relative timing

bjfletcher:
doing the date summarisation in the user land instead of server, to take into account user's own clock rather than server's


andygnewman:
@bjfletcher great suggestions, thanks.

andygnewman:
Code looks good and reasonably straightforward for a complex task. 👍 

I did have a quick conversation with Max about dates yesterday. Because if we do the summary date according to user time, what happens to all the other dates pertaining to that article - date on card and date on article, that I believe are server side dates?

So you should probably catch up with him and maybe demonstrate some examples with a different client side timezone to see what the differences will be.

Also when we did the summary dates server side we had test coverage for the logic, shame to see our test coverage go down :(

Especially as there was a use case I recently thought through and wanted to see how it worked - having an ad container in the stream list that only gets rendered at certain breakpoints or responsive ad unit only. Does the summary date appear against the article below the undisplayed ad unit or not.

Ideally it wouldn't show a summary date when the ad unit is not displayed, but should show summary date when the ad unit is displayed.

However that could be reasonably complex to code up the logic.

bjfletcher:
Hi Andy,

I've just written & added a new unit test for this. :)

Regarding server-side rendering, I typed about that in channel which I'll just copy and paste in here:

> You don’t see it, but server renders London time and makes it invisible, on the client side, if there’s no JS, user sees this London time; if there is JS, code replaces London time with user’s relative time.

> With the date summaries, if there’s no JS, nothing is shown and the user can look at the card London times; if there is JS, code puts up “Today”, “Yesterday”, and dates before that, relevant to user’s clock, at appropriate positions in the stream list.

> I tested that to its bones yesterday and it was all looking & working really great.

bjfletcher:
This is dependant on this PR going into `o-date`: https://github.com/Financial-Times/o-date/pull/57

PR #1206, opened 2016-07-07
Stream Pages: Company (market data) subheader

andygnewman:
Trello Card: https://trello.com/c/dFFBvOAj

Key To Dos:
- [x] Hit screen endpoint with dataPackages of pricePerformance and quote
- [x] Format the new data correctly
- [x] Render in existing subheader

Some examples:

UK Exchange - changed GBp (pence) to more common GBX, but ensured Market Cap is in GBP (pounds).
screen shot 2016-07-08 at 12 31 33

EUR
screen shot 2016-07-08 at 12 31 08

Volkswagen is an interesting case.
There is a now known data issue that means the API is passing us the incorrect stock symbol for VW (may happen in other cases also).
Whilst the incorrect stock symbol returned data from the quote endpoint we used to use, it doesn't return data for the screen endpoint we are now using to source more data.
In these cases, we do not display any market data.
screen shot 2016-07-08 at 12 32 10

USD
screen shot 2016-07-08 at 12 32 58


server/transforms/markets-data.js
```diff
@@ -1,50 +1,78 @@
...
 
-function formatMarketsData(item) {
+const marketCapCalculation = (value, currency) => {
+	const marketCap = {
+		value,
+		currency
+	};
+	// translates GBX (pence) to GBP (pounds) for market cap
+	if (currency === 'GBX') {
+		marketCap.value = value/100;
```
bjfletcher: you ok to not round up to 2 decimal points? e.g., what if it is 1.33333...?

bjfletcher:
Looks good to me :) :+1:

server/transforms/markets-data.js
andygnewman: gets rounded to 2 decimal places from line 36 when the overall formatting gets applies.

PR #1208, opened 2016-07-08
Repair: 'Related to your search' toggle (search results page)

andygout:
cc @andygnewman 

##### From:
screen shot 2016-07-08 at 09 55 13

##### To (with repaired toggle):
screen shot 2016-07-08 at 09 55 25


ironsidevsquincy:
Don't you just need to include the `o-expander` styles in the search page CSS?

andygout:
Am I right in thinking that this fix is required in relation to changes made here: https://github.com/Financial-Times/n-ui/commit/b8a9ee320c499ced6125bae5a404dbd456cef4a7#diff-02637b46ebfc732031e2457081a8bba7 ?

andygout:
Also - if I need to use these styles elsewhere (i.e. in both `sub-header` and `new-sub-header`), presumably I should just import them the once in a more centralised SCSS file (such as `main-stream.scss`)?

andygout:
Solution proposed in https://github.com/Financial-Times/n-ui/pull/218

PR #1209, opened 2016-07-08
robustify related special reports endpoint

andygnewman:
cc: @bjfletcher 

To solve the issue with related special reports.


views/related/special-reports.html
```diff
@@ -3,11 +3,15 @@
...
-					</div>
-				{{/each}}
+				{{#if specialReports}}
+					{{#each specialReports}}
+						<div class="hot-topics__column">
+							{{>special-report-item}}
+						</div>
+					{{/each}}
+				{{else}}
+					<p>No related special reports found</p>
```
bjfletcher: Is Max cool with this? Probably best to continue and throw error which will cause the CDN to use old, cached version.

andygnewman: It was probably needed anyway as there could be a legitimate instance where there are no special reports.

PR #1211, opened 2016-07-08
Style brands subheader

andygout:
cc @andygnewman @bjfletcher 

- [x] Review w/ Patricia; might be tweaks following that.

N.B. Uses existing `.sub-header` classes which are overwritten where needs be (rather than a whole new set of `.new-sub-header` classes (apart a couple of unavoidable instances) that would mostly replicate what the former are already doing) - let me know if you think wholly new set of classes is required.

##### 400px (within default) - toggle closed
400 closed

##### 400px (within default) - toggle open
400 open


client/components/new-sub-header/main.js
```diff
@@ -0,0 +1,29 @@
...
+		expandedToggleText: 'Show less'
+	});
+}
+
+module.exports = {
+
+	init: function (flags) {
+		if (flags.get('streamNewSubheader')) {
+			newSubHeaderToggle();
+		} else {
```
andygnewman: I was looking again at this (after we discussed it last night) and I don't think that you need the else.

andygout:
Roger that - will get on it.

client/components/new-sub-header/_main.scss
```diff
@@ -1,17 +1,126 @@
...
 }
 
-.new-sub-header__title {
+.new-sub-header__section {
+	position: static;
+	padding-left: 20px;
+
+	@include oGridRespondTo(M) {
+		padding-left: 16px;
+		padding: 0 10px 0 16px;
```
bjfletcher: `padding-left` above now redundant

client/components/new-sub-header/_main.scss
```diff
@@ -1,17 +1,126 @@
...
+
+.new-sub-header__tools__toggle {
+	font-size: 18px;
+	color: getColor('grey-tint4');
+
+	i {
+		@include oIconsGetIcon('arrow-down', getColor('grey-tint4'), 24, $apply-base-styles: false, $apply-width-height: false);
+	}
+
+	&[aria-expanded=true] {
```
bjfletcher: is there a class selector we can use instead of `aria-expanded`?

client/components/new-sub-header/main.js
```diff
@@ -0,0 +1,25 @@
+'use strict';
```
bjfletcher: eslint with your text editor? it'll complain that `'use strict';` is unnecessary in module

client/components/new-sub-header/_main.scss
```diff
@@ -1,17 +1,126 @@
 .new-sub-header {
+	padding: 15px 0;
 
-}
+	&.new-sub-header__brand {
```
andygnewman: I feel that brand is a modifier rather than an element, so class should be `new-sub-header--brand`

client/components/new-sub-header/main.js
```diff
@@ -0,0 +1,25 @@
...
+	Expander.init(target, {
+		toggleState: 'all',
+		shrinkTo: 'hidden',
+		toggleSelector: '.new-sub-header__tools__toggle',
+		collapsedToggleText: 'Show summary',
+		expandedToggleText: 'Show less'
+	});
+}
+
+module.exports = {
```
bjfletcher: can be consistent with the imports and use export here... `export function init(flags)...` or `export const init = (flags) => ...`

views/partials/new-sub-header/main.html
```diff
@@ -1,24 +1,51 @@
-<div role="navigation" aria-label="category navigation" class="new-sub-header {{#if subHeader}}new-sub-header__{{subHeader.type}}{{/if}}" data-trackable="sub-header">
+<div role="navigation" aria-label="category navigation" class="new-sub-header{{#if subHeader}} new-sub-header__{{subHeader.type}}{{/if}}" data-trackable="sub-header">
```
bjfletcher: maybe break this long line up, so I don't have to scroll to the right on GitHub ;)

views/partials/new-sub-header/description.html
```diff
@@ -0,0 +1,3 @@
+{{#if descriptionXML}}
```
andygnewman: I have a personal preference for keeping the logic about whether to display a partial in the main template, so the partial  becomes an 'absolute' thing and the template can be more easily understood.

views/partials/new-sub-header/main.html
```diff
@@ -1,24 +1,51 @@
...
 
 	<div class="o-grid-container">
```
bjfletcher: maybe merge this `div` with the parent here, reduce nesting by one level

views/partials/new-sub-header/main.html
```diff
@@ -1,24 +1,51 @@
...
+					{{/if}}
+				{{/with}}
+
+				{{#ifAll @root.flags.myFtApi conceptId}}
+					{{>n-ui/myft/templates/follow version="3" name=title}}
+				{{/ifAll}}
+
+				{{#with subHeader}}
+
+					{{#ifSome strapline descriptionXML}}
```
andygnewman: not sure I fully understand the need for repetition here in and out of the o-expander?

client/components/new-sub-header/_main.scss
andygout: There does not seem to be an option available in (o-expander)[https://github.com/Financial-Times/o-expander] to toggle a class name, only the button's text and `aria-expanded` state, so restrained by that. Maybe I can raise an issue requesting that option (as presumably using ARIA attributes for styling is not best practice)?

views/partials/new-sub-header/main.html
andygout: Took my lead from the search results which does the same thing: https://next.ft.com/search?q=westminster

It seemed the simplest thing to do in terms of extracting the text for wide viewport widths away from the `o-expander` logic (which hides content when page opens as required for narrower viewport widths).

client/components/new-sub-header/main.js
andygout: The below works but get linting error: `Parsing error: 'export' can not be used as shorthand property`.

```
export const init = flags => {
	if (flags.get('streamNewSubheader')) newSubHeaderToggle();
};
```

andygout: Hang on - this syntax works with no linting errors - will push this now:

```
export function init (flags) {
	if (flags.get('streamNewSubheader')) newSubHeaderToggle();
}
```

ironsidevsquincy: Y no fat arrow function..?

andygout: Weird - now I'm not getting a linting error with the former of those two. Changing now.

ironsidevsquincy: Any reason for it not to be the `default` export too?

If linting's being obtuse,

```
const init = flags => {
   ...
}

export default init;
```

views/partials/new-sub-header/main.html
andygnewman: Hmmm, not one to worry about now, but worth keeping in the back of our minds. Must be a better solution...

bjfletcher:
Regarding fixing the `aria-expanded` for styling issue, if it'd take you half a day to try and get this addressed, I wouldn't worry - there's a lot of tech debt that is much bigger than this. :)

Apart from AN's concern there... LGTM :+1:

PR #1213, opened 2016-07-11
Wider trending column

bjfletcher:
currently we have 9 columns for top stories, 3 for trending topics
we're going to change to 8 and 4 respectively


andygnewman:
Yep, should be as easy as that. May want to check that the image sizes vs srcset requirements are still optimised.

bjfletcher:
@andygnewman how'd I verify that they're optimal?
your knowledge on this could perhaps go on https://github.com/Financial-Times/n-image/blob/master/README.md

andygout:
I did some work on `n-image` so can help with this. Am looking through changes now.

andygout:
The array of values given as the third argument in the [layouts](https://github.com/Financial-Times/next-stream-page/blob/1767a4212f55177c02afe1cf57a8a2965830f6ba/config/layouts/lead-story.js#L18) uses the [createImageConfig](https://github.com/Financial-Times/next-stream-page/blob/master/config/layouts/helpers/create-image-config.js) method, which passes those values to [createImageSizes](https://github.com/Financial-Times/next-stream-page/blob/master/config/layouts/helpers/create-image-sizes.js), which sets the values as ultimately used by `n-image`.

In the layout files we just need to make sure that the image widths required at different viewport sizes are the minimum necessary. I'd always just set the images to maximum size and then manually move the viewport to each breakpoint to see what size the image was for each one respectively.

bjfletcher:
Andy G & I have tweaked & optimised the widths. It's looking good.

PR #1219, opened 2016-07-12
Ads 937

andrewgeorgiou1981:
@adgad @VladDubrovskis @andygnewman


server/lib/render-stream.js
```diff
@@ -73,12 +73,25 @@ module.exports = function (req, res) {
 				templateData.AdMidBothSmall = AdMidBothSmall;
 			} else {
 				if (templateData.stream && templateData.stream.items) {
-					templateData.stream.items.splice(6, 0, {
-						template: 'ad-mid-both-small'
-					});
+						if (templateData.cssVariant ==='live-opening' && res.locals.flags.adsFastFtStream){
+
+							for (let i=1;i<65;i++) {
```
adgad: This can be `i<templateData.stream.items.length`

adgad: Also maybe worth a comment to make it clear why we're starting from 1 and not 0

shared/components/ad-mid-instream/index.js
```diff
@@ -0,0 +1,31 @@
+const React = require('react');
+
+
+let nextSlotId = 0;
```
adgad: Is this going to be shared by all requests? i.e. if I hit the page, it will increment to 5, then when someone else comes along and hits the page it will increment to 10? I don't _think_ in this case it will be a problem since it's just an identifier - but maybe safer to pass in the number to this component as a paramerter?

adgad:
👍

server/lib/render-stream.js
```diff
@@ -73,12 +73,24 @@ module.exports = function (req, res) {
 				templateData.AdMidBothSmall = AdMidBothSmall;
 			} else {
 				if (templateData.stream && templateData.stream.items) {
-					templateData.stream.items.splice(6, 0, {
-						template: 'ad-mid-both-small'
-					});
+						if (templateData.cssVariant ==='live-opening' && res.locals.flags.adsFastFtStream){
+
+							for (let i=1;i<templateData.stream.items.length;i++) {
```
andygnewman: Am interested in your use of for rather than the more generally used map() here. Is this related to the mutation of the array that is being iterated over?

shared/components/stream-item/index.js
```diff
@@ -26,6 +27,9 @@ module.exports = (props) => {
 		case 'ad-mid-both-small':
 			child = <AdMidBothSmall />;
 			break;
+		case 'ad-mid-instream':
+				child = <AdMidInstream />;
```
andygnewman: Just a bit too much indentation.

server/lib/render-stream.js
andrewgeorgiou1981: So the idea was just to loop through the array and insert an ad (via the splice on the following lines) after every fifth stream-item. I think i see what you mean with using map; however wouldnt this result in a new array, rather than allowing us to splice into the existing array? Perhaps i've missed something here though.

ironsidevsquincy: `reduce`?

andygnewman: Whilst map() returns a new array, you wouldn't have to return it.  Instead you would do;
```
templateData.stream.items.map((item, index) => {
    if ((index+1)%6 ===0) {
        templateData.stream.items.splice(index, 0, {
            template: 'ad-mid-instream',
            id: i
        });
    }
});
```

andygnewman: Though interestingly I just tested this out and it doesn't take into account the increasing length of the array - so cuts out before the end. Hence I'm liking Darren's idea of reduce (though yet to figure out exactly how that would work).

ironsidevsquincy: If not interested in the return value, might as well just use `forEach`. But if you want to keep it functional

```
const newItems = templateData.stream.items.reduce((currentNewItems, item, index) => {
    if (index % 6 === 0) {
        currentNewItems.push({ template: 'ad-mid-instream', id: index });
    }
    currentNewItems.push(item);
    return currentNewItems;
}, []);
```

andygnewman:
👍

PR #1220, opened 2016-07-12
New special reports

bjfletcher:
Test coverage is 100% in all categories for the special report transform 🎉


server/lib/special-report-top-stories.js
```diff
@@ -0,0 +1,28 @@
+import api from 'next-ft-api-client';
+import fetchApiV2 from './fetch-api-v2';
+import concordance from './concordance';
+import articleModel from 'ft-n-content-model';
+
+// get the special report top story IDs
+function getTopStoryIds (streamId) {
```
ironsidevsquincy: ES6 arrow function?

server/lib/special-report-top-stories.js
```diff
@@ -0,0 +1,28 @@
...
+	return fetchApiV2({ path: `/lists?curatedSpecialReportsFor=${streamId}` })
+		.then(list => list && list.items ? list.items : [])
+		.then(list => list.map(item => item.id.split('/').pop()));
+}
+
+module.exports = (streamId) => {
```
ironsidevsquincy: seeing as your importing above as ES6 modules, for consistency would be good to export as an ES6 module too

server/transforms/special-report.js
```diff
@@ -1,13 +1,47 @@
-'use strict';
+import getSpecialReportTopStories from '../lib/special-report-top-stories';
+import getSection from '../../config/sections';
 
 module.exports = () => {
```
ironsidevsquincy: Ditto

server/lib/special-report-top-stories.js
bjfletcher: `const getTopStoryIds = (streamId) => {` you mean?

ironsidevsquincy: yeah. you can even omit the parenthesis if there's exactly one argument

```
const getTopStoryIds = streamId => {
```

ironsidevsquincy: Oh, and as there's only a single return statement in the body, you make it even shorter (https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Functions/Arrow_functions#Shorter_functions)

```
const getTopStoryIds = streamId =>
    fetchApiV2({ path: `/lists?curatedSpecialReportsFor=${streamId}` })
        .then(list => list && list.items ? list.items : [])
        .then(list => list.map(item => item.id.split('/').pop()));
````

File: server/lib/special-report-top-stories.js [hunk 4]
```diff
@@ -0,0 +1,28 @@
...
+	// get the V2 type rather than TME of the ID
+	return concordance.tmeToV2(streamId)
```
ironsidevsquincy: could use short function syntax here too

server/transforms/special-report.js
```diff
@@ -1,13 +1,47 @@
...
+
 	return class {
```
ironsidevsquincy: short function syntax?

server/lib/special-report-top-stories.js
bjfletcher: :+1: removing the need to return

Personally I find:

`function foo(blah) { ...`

more readable than:

`const foo = blah => { ...`

I'll be very interested to know why you like the latter. I'm open minded here :)

ironsidevsquincy: Well mainly because 99% of the time the arrow functions lexical `this` is far preferable to the 'old' function's `this`, so it's more a case of habit to always use arrow functions unless that 1% pops up and you need the old function's `this`

server/transforms/special-report.js
```diff
@@ -1,13 +1,50 @@
...
+					return this.specialReportTopStories.then(topStories => {
+						if (topStories.length > 0) {
+							const splitPoint = 1 + Math.ceil((topStories.length - 1) / 2);
+							const content = {
+								leadStory: topStories.slice(0, 1),
+								columnOne: topStories.slice(1, splitPoint),
+								columnTwo: topStories.slice(splitPoint)
+							};
+							// don't show related links for lead story
+							content.leadStory[0].relatedContent = [];
```
andygnewman: Does this need setting to an empty array here? Presume it would otherwise be undefined - the template should be able to handle undefined.

andygnewman:
Looks good. Just the one small comment on the empty array.

server/transforms/special-report.js
bjfletcher: @andygnewman `relatedContent` needs to be defined because we use the LeadStory layout config and it says that it has related links. We could either:

a) use empty array; or
b) create a new LeadStory layout config, say, LeadStorySpecialReport, with the related links flag turned off

I thought that adding whole new config was maybe a bit overkill when we can simply say that there aren't any related links by using an empty list.  I don't mind which approach we go for. Would you like to make the call here? :)

andygnewman: Agree it shouldn't need a new config, but the layout seems a bit brittle if it can not cope with a piece of content being undefined.
Would rather the layout didn't break if the data was undefined.

Issue #1214, opened 2016-07-11
Rename 'Latest news in' to 'Latest in'


mmockett:
https://github.com/Financial-Times/next-stream-page/blob/5fbfb343208f102c37ebde40817a58c91501a549/views/article-list.html#L136

mmockett:
Also should inherit the logic from More Ons: 
Latest in (section) / from (Author / Brand) / on (topic)

PR #1225, opened 2016-07-14
add prepositions and correct labelling to latest

andygnewman:
This Trello Card: https://trello.com/c/wCmCV9kF
This Issue: https://github.com/Financial-Times/next-stream-page/issues/1214

Still awaiting result on how to label 'Most Read'


config/sections/most-popular.js
```diff
@@ -1,6 +1,6 @@
 export default (name) => ({
 	id: 'most-popular',
-	title: `Most Read in ${name}`,
+	title: `Most Read${name.preposition} ${name.title}`,
```
bjfletcher: where's the space between 'Most Read' and `${name.preposition}`?

andygnewman: The preposition works out whether to put a space in front as sometimes the preposition will be blank.
https://github.com/Financial-Times/n-tag-preposition/blob/master/README.md

shared/components/stream-new/index.js
```diff
@@ -47,10 +47,22 @@ class StreamNew extends React.Component {
...
 
 		if (this.props.hasStreamListTitle) {
+			let titleContent;
+			switch (this.props.streamTaxonomy) {
+				case 'specialReports':
+					titleContent = 'More from this Special Report';
+					break;
+				case 'genre':
+					titleContent = 'Latest';
+					break;
```
andygout: Does this make [this line](https://github.com/Financial-Times/n-tag-preposition/blob/master/main.js#L17) in `n-tag-preposition` obsolete. Surely this would be the scenario in which it should be required?

andygout:
Other than comment 👍

shared/components/stream-new/index.js
andygnewman: Not entirely as there is a use case in article that uses it.

PR #1229, opened 2016-07-14
Remove: currentNav transform from Brands

andygout:
cc @andygnewman @i-like-robots @aintgoin2goa 

Adding `currentNav` transform to Brands is causing it to have empty breadcrumb area (owing to this line in [`n-ui/header/template.html`](https://github.com/Financial-Times/n-ui/blob/c0164404facf3dad660acc575ead78aec168ddce/header/template.html#L17)).

Saw it was explicitly added in this [PR](https://github.com/Financial-Times/next-stream-page/commit/82f967e78b75422a4472c7753f2fb68d944094a3) so want to be sure that there is no other reason why Brands might still need this transform?

screen shot 2016-07-14 at 16 31 53


i-like-robots:
You'll probably get in trouble now for it being missing on other brand pages! @mmockett - you made me add this originally, did you not?

I think the real question is why is `currentNav` anything other than `undefined` when the westminster blog does not appear in the [taxonomy](https://github.com/Financial-Times/n-navigation/blob/master/config/taxonomy.yaml).

i-like-robots:
https://financialtimes.slack.com/archives/next-navigation/p1464701084000387

andygout:
Addressing in this [PR](https://github.com/Financial-Times/next-stream-page/pull/1230) instead.

PR #1230, opened 2016-07-14
Amend: currentNav transform to make conditional on presence of navigation.currentItem

andygout:
cc @andygnewman @i-like-robots @aintgoin2goa 

Works on:-
`Opinion/FT View`
`Opinion/Lex`

Closing previous [PR](https://github.com/Financial-Times/next-stream-page/pull/1229) (which did not display breadcrumbs on above branded streams).


i-like-robots:
That's the ticket!

PR #1231, opened 2016-07-14
new ad rules & improved styling

bjfletcher:
Essentially:

- to show ad in place of trending topics for Business Education and Special Reports
- if ad loaded, show trending topics to right of the stream list
- if no ad loaded, show trending topics next to top stories
- all only in L and XL breakpoints

Outstanding works to do:

- [x] ask @adgad why the dummy ad doesn't have the loaded attribute set
- [x] once the above is resolved, then hide trending topics next to stream list *only* if dummy ad's loaded attribute is non-false
- [x] styling issue on Business Education page - "Latest..." and "Trending" don't align and has too wide a gap in between (see below)
- [x] styling issue on http://local.ft.com:5050/companies/banks - ad and "Trending" too close to each other and misaligned (see further below)
- [x] do testing in M, S and XS breakpoints, to check that styling is perfect in these breakpoints across all kinds of stream pages

Styling issue 1 (misaligned & too wide gap):

image

Styling issue 2 (misaligned & too close to each other):

image


andygnewman:
@bjfletcher I had a conversation with @adgad about this on the morning of the 14th. 

We threw around the idea of;
- rendering trending topics (hidden initially) in the top half right hand slot.
- use an event listener to find out if an ad has been loaded.
- if there is an ad in the side bar then move the trending topics to an empty div alongside the list.
- if there is no ad in the sidebar, unhide trending topics in the sidebar location.

In this was we would avoid double rendering of trending topics.

bjfletcher:
Hi @andygnewman, I had a chat with Arjun about that too, we felt that the JS route was a bit overkill... client-side JS-based rendering has the issue in that it has to wait until the JS is loaded and run.  Not great for performance.

The CSS approach is faster & doesn't need JS.

There's just one trending topic markup in most cases, only on Business Education page & Special Reports, that it is repeated twice... although only one will be rendered, the other will be behind `display: none`.

bjfletcher:
Stream and trending topics better aligned and not a bit too far from each other.

image

bjfletcher:
Better spacing between ad and trending topics. Note that we're going (in a separate PR) to switch to a 300px width for the right-hand column, so that ad fits in nicely and to give top section more room. 

image

andygnewman:
LGTM - build failure I presume is caused by version of nUi - a rebuild without cache should hopefully fix that.

bjfletcher:
styled the summary dates too and enforced 300px width for the right column in L and XL

PR #1238, opened 2016-07-18
[WIP] initial l&a routing

andygnewman:
Life and Arts Hub Page:

In this PR;

- [x] routing (based on flag toggle, streamLifeAndArts) that directs requests to a new 'hub' page controller
- [x] config file that drives the specific sub-topics for display on the hub page (and their order)
- [x] hub page controller that marshals the appropriate article data to populate the page, delegating this to various transforms;
    - [x] top-section-hub: to get lead story, subsequent 4 top stories and columnist articles
    - [x] most-read - reusing existing most read for top 6 most read in L&A category
    - [x] hub-topics - gets the first three curated top stories (or as a fallback, the 3 most recent stories) per sub-category


config/hub-pages.json
```diff
@@ -0,0 +1,50 @@
+{
+  "pages": [
```
matthew-andrews: ahh this looks familiar.  remind me tomorrow to dig out the web app code that does almost the same thing.  might save you a few minutes (-:

andygnewman: Thanks - would be good to see.

server/init.js
```diff
@@ -51,6 +53,14 @@ app.use(blogsPermalink);
 
 app.use(express.cacheMiddleware('short'));
 
+app.get('/stream/sectionsId/:thingId', (req, res, next) => {
```
matthew-andrews: :wink:

PR #1249, opened 2016-07-21
Life and Arts: Eight Top Stories

andygnewman:
cc: @andygout @bjfletcher 

screen shot 2016-07-21 at 12 18 33

Took the opportunity to also create a separate CSS bundle for hub page so as we're not loading unnecessary CSS eg for stream list, video highlights component etc. and can add in any hub page specific layout.


andygout:
Am guessing the lack of card borders in the screengrab is owing to the side effect of shrinking the viewport size in order to get a picture of the entire top section?

andygnewman:
Spot on, sounds like you've been there, done that, got the t-shirt.

PR #1253, opened 2016-07-21
Fix: Author pages (check for author.memberships before assigning strapline value)

andygout:
cc @andygnewman


andygnewman:
Base code changes 👍 , but make sure tests are passing before pushing live.

PR #1259, opened 2016-07-22
make the rss json optional

andygnewman:
cc: @matthew-andrews 

had broken our version behind the flag


matthew-andrews:
oops

matthew-andrews:
+1

PR #1265, opened 2016-07-25
Remove: Image conditionals from text show logic in author subheader

andygout:
cc @andygnewman @bjfletcher 

Authors who do not have an image will now nevertheless display all other information (strapline, description, Twitter handle, email address).


views/partials/new-sub-header/strapline.html
```diff
@@ -1 +1,3 @@
-<h2 class="new-sub-header__strapline">{{strapline}}</h2>
+<h2 class="new-sub-header__strapline{{#unless _imageUrl}} new-sub-header__strapline--without-img{{/unless}}">
```
andygnewman: Wondering why you're using a negative attribute - 'without' rather than a positive 'has' attribute in your logic. I think I know the answer, just feels a little weird.

andygout: I agree - likely because it's based on the absence of something. Would something like `--has-no-img` or `--has-absent-img` be the best wording in this scenario?

andygnewman:
could perhaps use the :not CSS property? https://css-tricks.com/almanac/selectors/n/not/

client/components/new-sub-header/_main.scss
```diff
@@ -127,6 +127,10 @@ $author-img-default-width-offset: $author-img-default-width + 40;
 	min-height: 24px;
 	margin-bottom: 10px;
 
+	&.new-sub-header__strapline--has-absent-img {
```
ironsidevsquincy: does this need the `&`? won't this selector be specific enough on its own?

views/partials/new-sub-header/strapline.html
```diff
@@ -1 +1,3 @@
-<h2 class="new-sub-header__strapline">{{strapline}}</h2>
+<h2 class="new-sub-header__strapline{{#unless _imageUrl}} new-sub-header__strapline--has-absent-img{{/unless}}">
```
ironsidevsquincy: Why not add `new-sub-header__strapline--has-image` if there is an image? And swap the width css around?

ironsidevsquincy: Where does the image appear? In the strapline? If not, probably shouldn't be a modifier of this (apologies if I've got the wrong end of the stick here)

andygout: Yup - reckon that's the way to go, and modifying the parent `new-sub-header__section` div rather than sibling `new-sub-header__strapline` h2.

client/components/new-sub-header/_main.scss
```diff
@@ -34,10 +34,13 @@ $author-img-default-width-offset: $author-img-default-width + 40;
 }
 
 .new-sub-header--author {
-	.new-sub-header__strapline {
-		max-width: calc(100% - #{$author-img-default-width});
-		@include oGridRespondTo(M) {
-			max-width: 100%;
+	.new-sub-header__section--has-img {
```
ironsidevsquincy: does this need to live in `.new-sub-header--author `?

andygout: Think so - brands subheader uses the same layout and styling, and also has an image, but is able to employ `float` given the image is not absolutely positioned as it is with author images (that need to stick to the bottom of the container within the padding area).

ironsidevsquincy: Ah, ok. Three levels deep of long selectors isn't the end of the world here, but if this was used more often I'd say see if we can make a class specifically for this case

PR #1266, opened 2016-07-25
Fix: Author image overlapping follow button

andygout:
cc @andygnewman @bjfletcher 

In the absence of a strapline, the unexpanded toggle will have a larger top margin to prevent the author image from clashing with the 'Follow' button at default-M viewport widths.


andygnewman:
👍

PR #1272, opened 2016-07-26
Move BizEd MPU higher, and more mobile/tablet ad slots

adgad:
/cc @andrewgeorgiou1981 @andygnewman @bjfletcher 

* Move the Business Education MPU next to Top Stories on Tablet (consistent with other top level sections).

* On stream pages (where there is no leaderboard), add an ad in the 12th slot (so there is one in the 6th and one in the 12th)


andygnewman:
@adgad understand the changes to pos 12 ad slot.

Not quite understanding how the changes to config/layouts/top-stories-column-two.js related to the first objective?

adgad:
So I essentially just copied the div from opinion-and-analysis layout - sp using top-stories-column-two purely for data-o-ads-formats-medium="MediumRectangle,Responsive", and there's already another slot in top-stories-column-one for data-o-ads-formats-small and data-o-ads-formats-default

andygnewman:
Thanks for updating the spreadsheet 👍

PR #1273, opened 2016-07-27
Life and Arts: handle no lunch portrait image

andygnewman:
Was causing a failure because one article had no image (due to a capi failure).
This defends against that scenario.


andygout:
👍

PR #1274, opened 2016-07-27
Nest: Lead quad layout into rows

andygout:
cc @andygnewman @bjfletcher 

Fixes the second issue on https://trello.com/c/QAcyleLD/300-bug-l-a-ie9-broken-layout.

`o-grid`'s flex polyfill for IE9 uses `float` to fix the rows, so each individual row should be created as such (rather than a row with excess content that spills onto two rows). Given in this scenario the layout remains consistent across all breakpoints we can do this as per these changes.


andygnewman:
👍

PR #1276, opened 2016-07-27
Suppress: Video highlights in IE8

andygout:
cc @andygnewman @bjfletcher 

Fixes https://trello.com/c/JjSK6xwS/313-stream-ie8-stream-pages-suppress-video-title-on-video-section-on-each-stream-page


andygnewman:
👍

Issue #1280, opened 2016-07-28
Improve layout structure of stream list


andygnewman:
At L breakpoint, the original design was for 8 cols stream list, 4 cols for ad / trending topics.

To improve space for the stream list, some styling overrides were done to reduce the 4 cols down to an effective 316px.

However this caused some issues because the override was of equal specificity as the original styling and didn't always take precedence.

This was quickly overcome by adding an element selector to the override.

However this is not the best long term solution and should be readdressed.

bjfletcher:
I think the solution you have now put in (overriding grid instead of increasing specificity) is ok.

andygnewman:
My point was not related to the solution for improving the overriding, but the fact that if we're overriding the grid, we probably shouldn't be using the grid in the first place.

bjfletcher:
@andygnewman I think we should continue and use `o-grid` on the page, just for those two particular columns on L and XL need it a bit different. I think it's reasonable to use `o-grid` and override where appropriate.

PR #1281, opened 2016-07-28
increase specificity for css overriding grid

andygnewman:
cc: @bjfletcher The issue was that there was equal specificity for the grid and the override, which on the US election page reversed what was happening elsewhere.

Have put in an additional div selector, not the ideal solution, (a hack on top of a hack), but have also created an [issue](https://github.com/Financial-Times/next-stream-page/issues/1280) for us to address this layout better.


client/components/section/_main.scss
```diff
@@ -74,15 +74,15 @@
 }
 
 // have 300px right column (fits ad nicely) and the rest for the left column
-.stream-section__left-column {
+div.stream-section__left-column {
```
bjfletcher: How about:

`.stream-section__left-column[data-o-grid-colspan~=L8]`

or even:

`.stream-section__left-column[data-o-grid-colspan]`

or, for it to look less hacky:

```
// override grid
[data-o-grid-colspan] {
	.stream-section__left-column& {
```

andygnewman:
@bjfletcher good suggestion, thanks.

PR #1283, opened 2016-07-28
ADS-937: updating ad slot on tablet portrait request a MediumRectangl…

andrewgeorgiou1981:
updating ad slot on tablet portrait request a MediumRectangle rather than a Leaderboard @adgad @andygnewman


adgad:
👍

Issue #1284, opened 2016-07-28
US election stream page is showing Brexit logo


callumlocke:
Due to [this line](https://github.com/Financial-Times/next-stream-page/blob/23eadddb3bb50266b7a042744c23bbf404021de0/client/components/new-sub-header/_main.scss#L423) I think.

It should be using this logo: https://ig.ft.com/_other/us-election-2016-stream/logo.svg

callumlocke:
Problem is the SVG doesn't work through the image service, you just get a black square.

Switching off branding as a quick fix.

bjfletcher:
@callumlocke is this problem ongoing?
https://next.ft.com/us-election-2016 right?

bjfletcher:
@callumlocke I see the PR now. Cool, thanks for this issue and the PR. :+1: Hopefully someone in the team will have a look.

andygnewman:
@callumlocke Can this issue now be closed?

PR #1285, opened 2016-07-28
Disable broken branding for us-election-2016

callumlocke:
Just a quick fix for #1284 (but doesn't fix it fully)


matthew-andrews:
+1

PR #1287, opened 2016-07-29
Cards with content overlay on image

andygout:
cc @andygnewman @bjfletcher 

Relies on: https://github.com/Financial-Times/n-card/pull/140

See this [PR](https://github.com/Financial-Times/n-card/pull/140) in `n-card` for illustrative screengrabs of resultant use of these changes.


config/layouts/lead-main.js
```diff
@@ -10,7 +10,8 @@ export default [
 				components: [
 					{
 						type: 'Content',
-						size: 'medium',
+						size: 'large-top-story',
+						imageHasTextOverlay: true,
```
ironsidevsquincy: seems like this should live in the `image` object property

andygout: Ostensibly yes, but it is essentially all related to the card content and where it is positioned on the card (albeit directly on top of the image), rather than the image itself, which remains untouched when creating this effect.

Could a solution be different naming so as to stress that changes are to the content not the image (`contentLaysOnImage`)?

ironsidevsquincy: I'd say, from a user's perspective, this adds text on top of the image, regardless as to how that's implemented (moving text around so it lays over the image or whatever), so seems to make sense for it to be a property of image

Basically this is configuration, it doesn't care about implementation details

ironsidevsquincy:
👍

PR #1288, opened 2016-08-01
Handle: Non-content results when fetching hub content

andygout:
@andygnewman @bjfletcher 

Temporary fix for handling non-results (in this case 'Lunch With The FT' section), which otherwise results in `"Cannot read property 'filter' of undefined"` when trying to call `.filter` on `topicBucket.content` and later when calling `.slice` on `bucket.content`.


ironsidevsquincy:
So you're keeping results where there is no `results.length`?

andygout:
Yes. There is definitely a better way, especially given how misleading this is when reading the code.

If no `id` is returned from the concordance then it returns an empty array (although its `typeof` value is actually `Object`):
https://github.com/Financial-Times/next-stream-page/blob/1d093c3f697d6518b99c5fd4ae12ac5fbe3fc9ca/server/lib/hub-topic-content.js#L14

This array has a `length` value of `0` while if a result is returned as an intended Object with data then its `length` is `undefined`. There didn't seem to be any other obvious (and semantically more logical) differences between the two by which to filter.

Ultimately we need to address a bigger problem of how to retain the integrity of the layout so I guess the emphasis here is on temporary.

ironsidevsquincy:
Offf, horrible response format from the api there

andygnewman:
I think there was probably an underlying issue in the code in the hubTopics transform that resulted in no Lunch with the FT as that situation shouldn't occur. So I think we need to investigate that aswell, but you're also right that we need to protect the integrity of the page layout against content failure in any topic.

PR #1289, opened 2016-08-01
Provide data for subnav from navigation middleware

i-like-robots:
For future compatibility with the more general purpose subnav implemented here: https://github.com/Financial-Times/n-ui/pull/315


andygout:
👍 

Is it worth adding a test to see if the `subsections` values are as expected (rather than just checking its presence as a key)?

i-like-robots:
I would say so, but I'd also say that should be the responsibility of the middleware rather than this app.

PR #1290, opened 2016-08-02
Add: Travel top stories

andygout:
@andygnewman 

### Travel
travel

### World
world


andygnewman:
Would be better to add it into the top level feeds section below rather than have it's own specific matcher.

andygout:
We could have done until the removal of Trending Topics and Most Read, which are part of the Top Level Feeds, so will need its separate profile. But eventually all the profiles that appear on the Life & Arts page will share the same logic and therefore be able to share a profile that has all the requisite transforms.

andygnewman:
👍

PR #1291, opened 2016-08-02
Life & Arts: Lunch with FT - handle missing mainImage property

andygnewman:
a missing mainImage property encountered when swapping out the landscape image for portrait for Lunch with the FT caused the page to fail.

This is a quick fix to this, but recognise that this 'Lunch with FT' specific bit of code needs refactoring.


andygout:
Can Is it worth reverting this line to its former state: https://github.com/Financial-Times/next-stream-page/blob/master/server/transforms/hub-content.js#L43

As it was before this PR: https://github.com/Financial-Times/next-stream-page/pull/1288

PR #1292, opened 2016-08-02
Remove: Second image from Life & Arts Books components

andygout:
cc @andygnewman 

https://trello.com/c/VRNIve8J/318-don-t-show-second-image-on-books-component


andygnewman:
👍

PR #1299, opened 2016-08-03
Life arts ads

andygout:
@andygnewman @bjfletcher 

I realise that [views/partials/ads/mid-both-small.html](https://github.com/Financial-Times/next-stream-page/blob/life-arts-ads/views/partials/ads/mid-both-small.html) doubles up on what [shared/components/ad-mid-both-small/index.js](https://github.com/Financial-Times/next-stream-page/blob/19b6ced35a05545ae4fd6197db8e41172be71064/shared/components/ad-mid-both-small/index.js) is doing, so maybe when I refactor other instances of ads can be called as a shared HTML partial instead of a React component (or maybe other way round)?


views/hub.html
```diff
@@ -75,6 +75,23 @@
 	{{/with}}
 
 	<div
+		class="o-ads advert--top-small-medium"
```
andygnewman: Would be good to move these ads into partials to keep the overall hub html more of a structural template than detailed.

andygout: Will do this in a separate branch as can then deal with other `view` files at the same time.

PR #1300, opened 2016-08-03
Add: Top Stories to Life & Arts' subsections' respective stream pages

andygout:
cc @andygnewman @bjfletcher 

https://trello.com/c/k0M589ez/327-add-top-stories-to-stream-pages-of-sections-included-in-life-arts


andygnewman:
See you've left trending topics in - presume that once we have the ad rules around putting the DMPU in the right hand side if there's no Trending Topics, then we can tighten up the lax rules on when Trending Topics appear, which were kept loose to mitigate against empty space.

andygout:
Yup - on this [separate branch](https://github.com/Financial-Times/next-stream-page/tree/display-sidebar-ad-if-trending-topics-absent) I'm using the absence of Trending Topics to dictate whether or not to include a sidebar ad.

I guess this can still be used as a good basis on which to do so, even if we are then changing the actual logic of when Trending Topics show?

andygnewman:
Yes

andygout:
Have branched from this branch to work on this [branch](https://github.com/Financial-Times/next-stream-page/tree/use-life-arts-most-read-for-subsection-pages) which will use Life & Arts' Most Read data for each subsection page's Most Read section.

**So that branch should be merged into this one before we merge this to master**

PR #1303, opened 2016-08-04
Display: Sidebar ad if trending topics absent

andygout:
cc @andygnewman @bjfletcher


server/lib/render-stream.js
```diff
@@ -67,10 +67,17 @@ module.exports = function (req, res) {
...
 				templateData.StreamNew = StreamNew;
 			}
+
 			// adding 'old' stream react component as default
 			templateData.Stream = Stream;
-			// Business Education page and Special Reports want sidebar ad
-			templateData.sidebarAd = req.params.thingId === 'MTI2-U2VjdGlvbnM=' || templateData.isSpecialReport;
+
+			// Display sidebar ad if Business Education, Special Reports or no Trending Topics present
+			const displaySidebarAd =
```
andygnewman: You could assign the value directly to the property rather than assigned it to the const then assign the constant to the property.

PR #1313, opened 2016-08-05
Revert "Revert "Lazy load fastFT images""

wheresrhys:
Reverts Financial-Times/next-stream-page#1310


shared/components/card-open-in-stream/inline-article/index.js
```diff
@@ -16,23 +16,21 @@ module.exports = React.createClass({
 		};
 	},
 
-	componentDidMount() {
-		// TODO oVideo
+	update() {
 		const nVideo = require('n-video');
```
andygnewman: can you leave the TODO oVideo in?

PR #1316, opened 2016-08-08
Update: n-section and n-card versions in bower.json

andygout:
cc @andygnewman 

The changes in `n-section` and `n-card` that required a major version change related to server-side changes. Have also built and ran locally and all looks good.


andygnewman:
👍

PR #1319, opened 2016-08-08
Add: Life & Arts scroll depth tracking

andygout:
cc @andygnewman 

This will also activate scroll depth tracking on other stream pages which have elements with the `js-track-scroll-event` class (but just weren't making any use of it until now). However (taking World as an example), the scroll tracking picture is incomplete given its 'Most Read' and 'Latest' sections are not being tracked (although top and bottom elements which sandwich them do), so I'll add those on a [separate branch](https://github.com/Financial-Times/next-stream-page/pull/1320) and make sure it is all merged together so we can take today as a starting point.


andygnewman:
👍

client/components/scroll-depth/scroll-depth.js
```diff
@@ -0,0 +1,30 @@
...
+	const observer = new IntersectionObserver(
+		function (changes) {
+			intersectionCallback(this, changes);
+		}
+	);
+	[...document.querySelectorAll('.js-track-scroll-event')]
+		.forEach(component => {
+			observer.observe(component);
+		});
+}
```
ironsidevsquincy: Is this lifted from the front page? Would make sense for a single one to exist in n-ui, and both apps to share it

andygout: Yup - that was our next step but just going to get this going here in the meantime.

PR #1321, opened 2016-08-09
Always show mid halfpage slot on stream pages

adgad:
/cc @andygnewman @andygout


server/lib/render-stream.js
```diff
@@ -79,9 +79,10 @@ module.exports = function (req, res) {
 				templateData.trendingTopics.data.content.length;
 
 			templateData.sidebarAd =
-				req.params.thingId === 'MTI2-U2VjdGlvbnM=' ||
+				templateData.topSection &&
```
andygnewman: would suggest
```
req.params.thingId === 'MTI2-U2VjdGlvbnM=' ||
templateData.isSpecialReport ||
(templateData.topSection && !trendingTopicsContentPresent);
```

andygnewman:
👍

PR #1322, opened 2016-08-09
add hub page to smoke tests

andygnewman:
As uses different controller and views from other stream / section pages, should have it covered by a smoke test


andygout:
👍

PR #1326, opened 2016-08-10
add n-ui built public files to gitignore

andygnewman:
Plus reordering existing ones into alphabetical order so more easily comparable with what's actually built.


laurieboyes:
Is there a reason stream page doesn't just ignore the entire `/public/` folder?

andygnewman:
At the time I thought the build process checked explicitly for the presence in .gitignore of the public files.

Not sure if that still stands.

laurieboyes:
Ahhh yeah that would explain it, cool, thank you 👍

PR #1327, opened 2016-08-10
Add: Tests for Most Read transform using parent data

andygout:
cc @ironsidevsquincy @andygnewman


test/server/transforms/most-read.test.js
```diff
@@ -1,80 +1,221 @@
...
+	const MostRead = mostRead(
+		{
+			useParent : opts.useParent || null
+		},
+		{
+			fetchGraphqlData : deps.fetchGraphqlData || null,
+			mostReadQuery : deps.mostReadQuery || null,
+			contentModel : deps.contentModelStub || null,
+			getSection : deps.getSection || null
+		}
```
ironsidevsquincy: why not just pass in `opts` and `deps` directly?

PR #1328, opened 2016-08-11
Enable premium content indicator by passing flags through stream list to n-card

andygnewman:
Because we're bypassing n-section and going to n-card directly we need to pass through flags to enable toggling the function on and off for the stream list.


server/controllers/search.js
```diff
@@ -91,7 +91,8 @@ module.exports = function(req, res, next) {
 			const items = articles.length > 0
 				? articles.map(articleModel({
 						hasActions: res.locals.flags.myFtApi && res.locals.flags.myFtApiWrite,
-						newList: res.locals.flags.streamNewList
+						newList: res.locals.flags.streamNewList,
+						flags: res.locals.flags
```
ironsidevsquincy: previously we just passed in the option we needed based on a flag, rather than all the flags; any reason not to continue this?

andygnewman: Was maintaining parity with current utilisation of flags in n-card (and resolving existing live issue whereby the use of flags in tags wasn't working for stream page items). Usually all the flags would get sent through via n-section. For stream list we miss out n-section and go direct to n-card.

PR #1330, opened 2016-08-12
Remove n-video and replace with o-video

adgad:
/cc @VladDubrovskis @lc512k @andygnewman @andygnewman @wheresrhys 

Completely removes n-video and keeps o-video as default. Start using o-video for the inline card article (fastFT). Note: I've not been able to find a fastFT article that has a video in the first par - so that part is entirely guesswork.


VladDubrovskis:
👍

wheresrhys:
woop. Do you fancy putting o-video into n-ui now? I held off until we'd deprecated n-video

PR #1331, opened 2016-08-12
Amend: Business Education page does not render Trending Topics in Top Section

andygout:
cc @andygnewman 

https://trello.com/c/gQL1ZhrD/337-business-education-trending-topics-shows-if-no-ad

Have discovered that the issue of having two instances of Trending Topics arises mostly if refreshing when deeper down the page, caused by the lazy-loading logic (the rest of the time it seems to behave, unless the ad server simply doesn't return data), causing the element attribute `data-o-ads-loaded=false` to be applied.

Given the section to the right of the Top Stories *should* always render an ad, this seems to be the simplest fix. Ultimately, if ads can find a solution to the above them we can safely reinstate the code, but otherwise as is, it runs the risk of rendering Trending Topics twice.

We can remove the styling because it only ever applied to the Business Education scenario (the others were Special Reports, which don't seem to have a Top Section, and Top Sections without Trending Topics).


andygnewman:
Suggest we get rid of the dummy partial and dummy/trending-topics related CSS

PR #1333, opened 2016-08-16
Show FirstFt signup email component on Companies section page

ironsidevsquincy:
Also rename references to `light-signup` (that's the general term for this component)


andygnewman:
Always been a fan of skinny controllers, so wondering whether you can start using the 'transform' model that we have to do this, otherwise we're configuring different streams in different places and could get messy quite quickly.

Plus it makes test coverage, debugging a bit easier.

i-like-robots:
Details about authoring a transform are in the wiki https://github.com/Financial-Times/next-stream-page/wiki/transforms

ironsidevsquincy:
Will update

ironsidevsquincy:
Updated (love this transform stuff btw)

andygnewman:
Looking good. Just a couple of comments.

Putting the uuids in the transform rather than the profiles. I think this makes sense given that it's a very light transform and would add a lot more complexity to the profile config.

Using the transforms makes writing tests super easy (nudge, nudge).

ironsidevsquincy:
Lol. I thought you meant easy for somebody _else_. 

Ok, ok, I was being lazy, tests on their way

ironsidevsquincy:
Now with tests

test/server/transforms/light-signup.test.js
```diff
@@ -0,0 +1,67 @@
...
+			const res = {
+				templateData: { },
+				locals: { anon: { userIsAnonymous: false } }
+			};
+			const lightSignupTransform = new (lightSignup());
+			lightSignupTransform.showLightSignup = true;
+			lightSignupTransform.post(res, { firstFtEmailSignup: true });
+			res.templateData.showLightSignup.should.be.false;
+		});
+
```
andygnewman: Maybe the one missing scenario for completeness;
```
it('should not show if not enabled for this concept', () => {
    const res = {
        templateData: {},
        locals: { anon: { userIsAnonymous: true } }
    };
    const lightSignupTransform = new (lightSignup());
    lightSignupTransform.showLightSignup = false;
    lightSignupTransform.post(res, { firstFtEmailSignup: true });
    res.templateData.showLightSignup.should.be.false;
});
```

ironsidevsquincy: Good point

andygnewman:
👍  (sorry for making that more difficult, but worth it I think)

ironsidevsquincy:
No problem, might as well do it properly

PR #1334, opened 2016-08-16
Refactor render stream

andygnewman:
Objective: make the construction of presentational factors for render stream simpler to understand, easier to maintain and fully tested.

Key aspects;
- separate presentational elements into 4 key areas;
    components - gathering up the appropriate react components
    cssVariant - choosing the appropriate cssVariant
    ads - inserting ads into content lists and setting ad properties (any components in components)
    properties - complex calculations for booleans used in handlebars templates
 
- hold react components all into a specific `components` property of templateData
- hold calculated properties in a specific 'properties' property of templateData


andygout:
👍  and another 👍  for all the test coverage.

PR #1337, opened 2016-08-18
fix: Point RSS feed URLs to the correct domain name and protocol

matthew-andrews:
fyi @andygnewman @andygout @bjfletcher


andygnewman:
👍

PR #1341, opened 2016-08-18
Amend: Makefile to convert app to multi-region

andygout:
cc @matthew-andrews @andygnewman  @bjfletcher 

My understanding of what should be removed/added is in these [changes](https://github.com/Financial-Times/next/pull/128) I've made to the instructions (taken from reading through [`n-heroku-tools`](https://github.com/Financial-Times/n-heroku-tools) docs and code) and comparing to `next-front-page`'s [Makefile](https://github.com/Financial-Times/next-front-page/blob/master/Makefile) (given it is already a multi-region app)).

The build on this branch gives responses to the `float` command the same as those seen on `next-front-page` so expect I can take that to mean it is working correctly.

#### [Creating a new Multi-Region App (or converting an existing one)](http://financial-times.github.io/next/docs/developer-guide/multi-region-apps/)
- [x] 1. Install Dependencies
- [x] 2. Set up a pipeline
- [x] 3. Change your Makefile
- [ ] 4. Push and Hope!
- [ ] 5. Amend the service registry
- [ ] 6. Delete the old app

Set up a pipeline:
screen shot 2016-08-18 at 14 04 31


matthew-andrews:
lgtm

PR #1346, opened 2016-08-19
Improve related Special Report images

andygnewman:
- [x] Ensure there is always and image for each related special report
- [x] Bring the styling back (ie. align it with the changes made to photo diary)


andygout:
👍  post-linting fix.

PR #1351, opened 2016-08-22
mock live stream

VladDubrovskis:
@bjfletcher @andygnewman @i-like-robots 

Added this in order to locally mock the stories added in stream (FastFT) but this should work locally for any stream.

Using random page for stream and random number between 1 and 10 to get number of entries.

Done a PR to get some ideas if this is useful/worth wrapping up and if anyone has got any idea what should trigger it (what should go into "mockLiveStream"). A flag?


VladDubrovskis:
though the more I think about it the more it feels wrong to have any sort of "development" logic to be in master, even behind a flag. Will wait for some feedback and might close this. But leave for future reference :)

bjfletcher:
I think we should definitely at least keep this branch alive. Then anyone who wants to stimulate live stream, can refer to this branch. You'll save them/us some time. :)

Maybe make a little note in `README` referring to this branch.

If we can come up with fewer lines, then perhaps include it all in `master` but commented out. Then the developer can simply uncomment to simulate.

Just some ideas for discussion. :) What other ways have you seen elsewhere for doing this?

VladDubrovskis:
Have not seen any other ways but rather run with Matts idea of polling other streams. Then whilst playing around with it Arjun asked if we can poll same stream but different page that worked, so just made it a little more random.

Going to finish off my the story I am working on and see if I can simplify it or use less lines but think its quite minimal as it is

VladDubrovskis:
Closing this for now not to leave unfinished work laying around. Cannot keep ownership of but if anyone might need it might be a good idea to bookmark it :)

PR #1352, opened 2016-08-22
use v6 of node

bjfletcher:
up to the latest and greatest release of Node


package.json
```diff
@@ -102,6 +102,6 @@
     ]
   },
   "engines": {
-    "node": "4.4.2"
+    "node": ">=6.4.0"
```
ironsidevsquincy: It's annoying that you can use semver style in circle.yml, but I'd set this as exactly `6.4.0` (just so we know tests and app are running exactly the same versions)

andygnewman:
Can you let us know when you merge it in, I'd suggest that we then give it a couple of hours before merging anything else in, in case there are any performance / memory issues that arise out of putting it into prod.

package.json
bjfletcher: Done.

By the way... if all our Next apps use different minor versions of Node in v4 or v6 then we probably should have some kind of automatic Node version switching when we `cd` into a project, don't you think?

ironsidevsquincy: I uses https://github.com/Kikobeats/nodengine

PR #1354, opened 2016-08-23
Fix: Check all hub topic articles for matching tag

andygnewman:
Highlighted by an Arts top story used on the hub page that wasn't tagged with Arts.
The code's expectation was that each story in a top list would be tagged with the section's overall tag.

Have amended the code to search through all articles in the list for matching tag.
And if none found, then not try and populate the section name or link.


bjfletcher:
Nice one :+1:

PR #1355, opened 2016-08-23
removed use of res.cache() helper

wheresrhys:


andygnewman:
Looks good, even better if you fix the tests...

andygnewman:
Those tests are not valid - you're testing a value that you're setting in the test.

The best way to do it is as it was done before, set res.set as a sinon.stub returning null.

Then test that the stub is being called with the expected parameters..

wheresrhys:
they are valid. I set a dummy value as a property of res (mirroring https://github.com/Financial-Times/n-express/blob/master/src/middleware/cache.js#L138-L143), and then the test verifies that the transform has assigned them to the appropriate headers... or am I missing something?

PR #1360, opened 2016-08-31
Jobs box

VladDubrovskis:
/cc @andrewgeorgiou1981 @adgad 
screen shot 2016-08-31 at 15 14 43

Adds "Executive Appointments" box to the Work & Careers page.


shared/components/jobs-box/_main.scss
```diff
@@ -0,0 +1,88 @@
...
+	@include oTypographySerifDisplayBoldSize(l);
+	font-family: oFontsGetFontFamilyWithFallbacks(FinancierDisplayWeb);
+	position: absolute;
+	background-color: oColorsGetPaletteColor('pink');
+	color: oColorsGetPaletteColor('grey-tint4');
+	display: block;
+	margin: 0;
+	padding: 0 10px;
+	left: 10px;
+	top: -17px;
```
adgad: This seems like an odd thing to have to do?

test/server/transforms/jobs-box.test.js
```diff
@@ -0,0 +1,98 @@
+/*global describe, context, it, beforeEach */
+
+'use strict';
+
+const proxyquire = require('proxyquire');
+const expect = require('chai').expect;
+const requestResponse = require('./lib/request-response');
+
+function pollerStub() {}
```
VladDubrovskis: The stub is hardcoded here as the other poller-stub available seem to process the data that it gets. Also had to hard code the XML otherwise require() breaks the build/tests

server/config/profiles.js
```diff
@@ -81,6 +82,18 @@ module.exports = [
 			videoHighlights(),
 		]
 	},
+	{
+		name: 'Work & Careers - Jobs Box',
+		match: req => req.params.thingId === 'MTI1-U2VjdGlvbnM=',
```
VladDubrovskis: There is not test for this and could not find a test that checks certain transforms only happen on certain pages. Not sure if needs one.

andygnewman: We haven't cracked how to test the overall allocation of transforms to pages. Will need to put thinking caps on for that.

andygnewman: You can also remove the Work & Career entry from line 107.

server/transforms/jobs-box.js
```diff
@@ -0,0 +1,57 @@
...
+				});
+				reject(err);
+			}
+		});
+	});
+}
+
+module.exports = () => {
+	return class JobsBox {
+		post(res, flags) {
```
andygnewman: Should be able to kick some of this off in a pre state. See the non-blocking method in the wiki https://github.com/Financial-Times/next-stream-page/wiki/Transforms#authoring-transforms.

An example is here https://github.com/Financial-Times/next-stream-page/blob/master/server/transforms/columnists.js

shared/components/jobs-box/_main.scss
```diff
@@ -0,0 +1,80 @@
...
+	}
+}
+
+.jobs-box__search-form {
+	border-bottom: 1px solid oColorsGetPaletteColor('pink-tint3');
+}
+
+.jobs-box__title {
+	@include oTypographySerifDisplayBoldSize(l);
+	font-family: oFontsGetFontFamilyWithFallbacks(FinancierDisplayWeb);
```
andygnewman: not sure why you're declaring a font-family aswell as oTypography size mixin, should be able to do all in one.
You may also want to load progressively - see https://github.com/Financial-Times/o-typography#progressively-loading-web-fonts

shared/components/jobs-box/_main.scss
```diff
@@ -0,0 +1,80 @@
...
+		color: getColor('link-1');
+	}
+	strong {
+		font-weight: 600;
+	}
+}
+
+.jobs-box__newsletter-link {
+	@include oTypographySans(s);
+	font-weight: 600;
```
andygnewman: Should try and use the overall oTypography mixin that has the appropriate font weight.

shared/components/jobs-box/_main.scss
```diff
@@ -0,0 +1,80 @@
...
+	&:hover {
+		color: getColor('link-1');
+	}
+	strong {
+		font-weight: 600;
+	}
+}
+
+.jobs-box__newsletter-link {
+	@include oTypographySans(s);
```
andygnewman: a lot of repetition of oTypographySans(s) can it be set on an element higher up the tree and then altered where necessary?

shared/components/jobs-box/index.js
```diff
@@ -0,0 +1,41 @@
+'use strict';
+
+const React = require('react');
+
+module.exports = React.createClass({
```
andygnewman: given the simplicity of this react component it can be written functionally rather than class based.

eg. 
```
module.exports = (props) => {
    const items = props.items;
    return (
    );
};
```
Should have a slight performance benefit (probably not noticeable but best practice) (and yes I learnt that on the React course yesterday!)

server/lib/presentation/components.js
```diff
@@ -37,5 +38,9 @@ export default (templateData) => {
 		}
 	}
 
+	if (templateData.jobsBox) {
```
andygnewman: There's a series of tests for the components allocation, so you may want to add this in.

VladDubrovskis: Done.

andygnewman:
Nice 👍

VladDubrovskis:
thanks for reviewing and feedback @andygnewman

PR #1369, opened 2016-09-06
Rerender mid fastft ads if react overwrites dom

adgad:
@VladDubrovskis @andygnewman @andygout @wheresrhys 

The nUiBundle introduced a race condition whereby oAds would initialise the ads before react would rerender the fastFT stream - which meant that the in-stream fastFT ads wouldn't be there. This PR in conjunction with the below fixes that. _stops pulling hair out_

https://github.com/Financial-Times/n-ui/pull/402


shared/components/stream-new/index.js
```diff
@@ -18,6 +18,10 @@ class StreamNew extends React.Component {
 	}
 
 	componentDidMount () {
+		const adContainers = this.refs.list.querySelectorAll('.o-ads');
+		if(window.oAds && window.oAds.config('formats') && Object.keys(window.oAds.slots).length) {
```
wheresrhys: ` && flags.get('ads')` ?

adgad: Ooh good shout

adgad: Passing the flags down down not working, but made the check into something that will be false if the flag is off.

PR #1370, opened 2016-09-06
video styling not silent

andygnewman:


i-like-robots:
Sorry, I just came to fix this but you're too fast!

Because I did not tag the o-video release I thought we were safe but we're actually tracking a branch.

andygnewman:
No problem, managed to diagnose it, have done the article app as well.

PR #1376, opened 2016-09-09
Add styling for LSU on live and live-opening pages

ironsidevsquincy:


andygnewman:
👍  even though criticalCSS is getting quite large now :(

ironsidevsquincy:
Yeah, and this is only for anon users
Different CSS files for logged in/anon users? (this is getting quite complicated now)

PR #1377, opened 2016-09-12
Featured Special Reports for Special Reports Hub Page

andygnewman:
Objective:
- add a 'Featured' section to the top of the Special Reports hub page (https://www.ft.com/special-reports)

- This is a temporary solution until Special Reports are themselves items of content.
- Will work from a list of 'dummy stories' created by editorial for featured special reports which will have redirects to the featured special report.
- Behind a flag - streamSpecialReportsTopStories, currently running off World top stories until list populated for Special Reports

Example
screen shot 2016-09-12 at 14 22 58


server/controllers/special-reports.js
```diff
@@ -1,26 +1,38 @@
...
+	// Run request transforms
+	return transform.pre(req, flags)
 
-			return {
+		// Fetch special reports items
+		.then(() => fetchSpecialReports())
+
+		// Assemble basic response template data
+		.then(results => {
+			req.perPage.total = results.length;
```
bjfletcher: is this line intentional? (if it was me, could have been a debugging hack ;))

andygnewman: Yes was intentional, came across from L15 of original. I think the intention is for pagination to be able to use it to say 'page 3 of 137' but don't think we ever use it.

server/controllers/special-reports.js
```diff
@@ -33,13 +45,18 @@ module.exports = function(req, res, next) {
...
+		// Run response transforms
+		.then(() => transform.post(res, flags))
+
+		// Finally render the special reports hub list
+		.then(() => renderStream(req, res)(res.templateData))
+		.catch(err => {
+			logger.error(err);
+			throw err;
 		})
 		.catch(upstreamErrorHandler(res, next, 404));
```
bjfletcher: two catches here? merge into one perhaps

bjfletcher:
a few comments but looking really good :+1:

bjfletcher:
remove these smoke tests?

```
Man overboard http://ft-stream-branch-4398.herokuapp.com/special-reports.rss: bad status: 500
http://ft-stream-branch-4398.herokuapp.com/special-reports.rss: bad status: 500
```

andygnewman:
Those smoke tests are valid. I found that my match in config was only picking up the html version and not the rss / json ones, which were going to default transforms and therefore failing.

However we should stop doing any transforms for rss / json as they only return the stream list.

PR #1381, opened 2016-09-14
Swap out n-content-model for n-content-decorator

andygnewman:
n-content-decorator supersedes  n-content-model in that it only transforms or adds new properties whilst allowing the consuming app to retain all the existing data.

There are two usage options for n-content-decorator - either for it to mutate the content object on behalf of the consuming app, or to pass back an object with the new & transformed properties that the owning app can combine with the existing content object.

The implementation here is the latter option - the stream app combines the resulting properties with the existing ones.


server/lib/get-articles-for-list-type.js
```diff
@@ -1,7 +1,7 @@
 import api from 'next-ft-api-client';
 import fetchApiV2 from './fetch-api-v2';
 import concordance from './concordance';
-import contentModel from 'ft-n-content-model';
+import contentDecorator from 'ft-n-content-decorator';
```
bjfletcher: how about taking the opportunity here to switch to `@financial-times/n-content-decorator`?

bjfletcher:
:+1: (apart from the library name - @financial-times/content-decorator or @financial-times/n-content-decorator?)

server/lib/get-articles-for-list-type.js
andygnewman: Good call.

PR #1384, opened 2016-09-15
Update the source of the URL 

richard-still-ft:
i.e. use ES derived data not the incoming header.


server/controllers/hub.js
```diff
@@ -33,19 +33,17 @@ module.exports = (req, res, next) => {
...
 			},
 			// TODO: refactor this data!
-			path,
+			path: tag && `/stream/${tag.taxonomy}Id/${tag.id}`,
 			title: tag && tag.name,
 			taxonomy: tag && tag.taxonomy,
 			conceptId: tag && tag.id,
 			layout: 'hub',
-			vanityUrl,
+			vanityUrl: tag.url,
```
andygnewman: given all the other uses of tag here check for a tag (tag &&) would suggest doing that here. Though can't remember the reason why it is here and not in the stream controller.

i-like-robots: https://github.com/Financial-Times/next-stream-page/blob/d10ef039f83d680fa9d3de215d9703762aae4e64/server/controllers/stream.js#L46

andygnewman: Good spot, @richard-still-ft can you tidy this up as per the stream controller link posted above.

server/controllers/hub.js
```diff
@@ -32,20 +32,20 @@ module.exports = (req, res, next) => {
 
 	// Assemble basic response template data
 	.then((thingTag) => {
-		const tag = thingTag.items.shift();
-		const path = `/stream/${tag.taxonomy}Id/${tag.id}`;
-		const vanityUrl = req.headers['ft-vanity-url'] ? `https://www.ft.com${req.headers['ft-vanity-url']}` : `https://www.ft.com${path}`;
+		let tag = {};
+		tag = thingTag.items.shift();
```
andygnewman: could probably do it as
```
const tag = thingTag.items ? thingTag.items.shift() : {};
```
Otherwise you're always overwriting tag, possibly with undefined or actually 500ing if it can't shift() on undefined.

richard-still-ft: Good point will try to do it in less of a rush!

server/controllers/hub.js
```diff
@@ -32,8 +32,7 @@ module.exports = (req, res, next) => {
 
 	// Assemble basic response template data
 	.then((thingTag) => {
-		let tag = {};
-		tag = thingTag.items.shift();
+		let tag = thingTag.items ? thingTag.items.shift() : {};
```
andygnewman: const

PR #1386, opened 2016-09-19
add warn logging for bodyXML/bodyHTML usages

bjfletcher:
The motivation behind this is to speed up ES queries by excluding the heavy bodyXML/bodyHTML fields.  The idea is to add logging for usage of these fields, and then we can look on Splunk to understand when/where/if they are used.


andygout:
I've not had much direct experience with `n-logger` but in terms of semantics is there a function other than `warn` that could be used given these are more notifications rather than warnings.

andygnewman:
I'm not clear on the purpose of logging this, when it seems evident from the codebase where and when those fields are used?

I think we should focus on swapping out usage of the XML for the HTML versions, so at least we can drop one heavy field, along with the annotations.

bjfletcher:
@andygout our goal is to stop using those fields, and hence the "warn" level... a bit like with the CSS/JS linters

@andygnewman I tried the approach you suggested, but it depends on what fields are set from the CAPI/backend... it's not possible to tell from the codebase whether they are set or not. It's like this: `if (something is set on object from CAPI) { ... bodyHTML }`

The logging is just to help us understand it all better.  Unless you have all the answers? :)

andygout:
@bjfletcher Roger that, makes sense.

server/lib/lunch-with-ft-images.js
```diff
@@ -6,6 +7,7 @@ export default articles => {
 		const transformedArticle = Object.assign({}, article);
 		const $img = cheerio.load(transformedArticle.bodyXML)('img').first();
 		if ($img.length > 0) {
+			logger.warn({ event: 'BODY_XML_USAGE', id: article.id });
```
andygnewman: bodyXML is being used here to hack in the portrait Lunch with FT images as the mainImage properties rather than the landscape images. This is used exclusively for the Life & Arts hub page.
Could be done with bodyHTML as an alternative.

server/models/subheading.js
```diff
@@ -1,12 +1,15 @@
+const logger = require('@financial-times/n-logger').default;
 const cheerio = require('cheerio');
 
 module.exports = item => {
 
 	if (item.summaries && item.summaries.length) {
 		return item.summaries[0];
 	} else if (item.openingXML) {
```
andygnewman: This uses a hierarchical approach to populating summaries, should it be missing from the content, by using either the first p of the opening or the body.
Would be an idea to log how often these get triggered and whether this back up logic is needed.

server/transforms/open-in-stream.js
```diff
@@ -18,6 +19,7 @@ module.exports = () => {
 				res.templateData.stream.items.map((item, i) => {
 					// if no opening, use the first p of the body as the opening
 					if (!item.openingHTML && item.bodyHTML) {
+						logger.warn({ event: 'BODY_HTML_USAGE', id: item.id });
```
andygnewman: This usage is for open in stream (used on FastFT and Photo Diary).

Would be invoked mainly on Photo Diary as all FastFT articles will have an openingHTML.

server/lib/lunch-with-ft-images.js
bjfletcher: How are we going to eventually get rid of bodyHTML here?

server/models/subheading.js
bjfletcher: Line 9 (right below your comment) has the logging. Isn't that enough?

server/transforms/open-in-stream.js
bjfletcher: So, it's used only in Photo Diary?

server/lib/lunch-with-ft-images.js
andygnewman: No obvious solution.

server/transforms/open-in-stream.js
andygnewman: Probably

andygnewman:
We could focus on getting quite easily to the point whereby we are excluding the annotations, one of the bodies (HTML or XML) and the openingXML from the return from ES. Get that live and see what improvement we're making. Then we could look at the more difficult options for excluding the other body property.

server/models/subheading.js
bjfletcher: According to the logs, this backup logic has never been executed since 3pm yesterday.

PR #1389, opened 2016-09-20
Amend: Migrate stateless React components to functional basis

andygnewman:
cc: @i-like-robots 

A small selection (more to come)

- migrate components to functions
- incorporate PropTypes validations


i-like-robots:
LGTM

PR #1390, opened 2016-09-20
Optimise ElasticSearch

bjfletcher:
In local testing, excluding some heavy fields reduced response time significantly. So, let's do this!

This PR:

- exclude heavy fields from ES responses as follows:
  - if FastFT/Photo Diary: 'openingXML', 'bodyXML', 'annotations'
  - if Life & Arts: 'openingHTML', 'openingXML', 'bodyHTML', 'annotations'
  - all other pages: 'openingHTML', 'openingXML', 'bodyHTML', 'bodyXML', 'annotations'
- remove subheading module and test as they are dead code
- some "never fail" tests fixed

I tried switching Life & Arts to using HTML instead of XML, but had some "anonymous" handlebars issues I could not get to the bottom of. I'm not too worried because a) XML is probably faster than HTML to parse, and b) we're requesting XML for **only** Life & Arts.


server/controllers/stream.js
```diff
@@ -7,8 +7,23 @@ const profiles = require('../config/profiles');
 const logger = require('@financial-times/n-logger').default;
 const addTagPreposition = require('ft-n-tag-preposition');
 
+const getEsExcludes = (thingId) => {
+	const heavyFields = ['openingHTML', 'openingXML', 'bodyHTML', 'bodyXML', 'annotations'];
+	const isLifeAndArts = thingId === 'MTQ4-U2VjdGlvbnM=';
```
andygnewman: LifeAndArts avoids this controller (goes to hub controller) so I don't think that this is doing anything.

server/controllers/stream.js
```diff
@@ -7,8 +7,23 @@ const profiles = require('../config/profiles');
...
+	const isFastFt = thingId === 'NTlhNzEyMzMtZjBjZi00Y2U1LTg0ODUtZWVjNmEyYmU1NzQ2-QnJhbmRz';
```
andygnewman: as the requirement is to retain the openingHTML and bodyHTML is streams that have cards that openInStream, it would be better to use that attribute rather than the specific stream ids.

This would future proof it from any further additions and having to remember to update it here.

bjfletcher: @andygnewman that's tricky as `openInStream` is in the `post` transform, not `pre`. We need this in `pre` before feeding to ES.

How about if I move `openInStream` code into `pre` from `post`?

andygnewman: Yep that makes sense, there's nothing in the setting of it that isn't available in pre.

server/controllers/stream.js
andygnewman: This is actually where L&A gets it's content from;
https://github.com/Financial-Times/next-stream-page/blob/master/server/lib/hub-topic-content.js#L18

Uses api.content, based on there being a relevant list, otherwise (as a back up) it uses the api.search based on stream id. But small use case to worry about optimising.

andygnewman:
LGTM 👍

PR #1397, opened 2016-09-21
Update: n.Makefile and lintspaces-cli version

andygout:
cc @ironsidevsquincy @andygnewman @bjfletcher


server/lib/load-list.js
```diff
@@ -3,8 +3,8 @@ import api from 'next-ft-api-client';
 import {coroutine} from 'bluebird';
 
 /**
- * Downloads a list of articles from CAPI or Bertha and returns the UUIDs, plus any metadata.
- */
+* Downloads a list of articles from CAPI or Bertha and returns the UUIDs, plus any metadata.
+*/
```
ironsidevsquincy: hmmmm, these should be ok...what's the error?

andygout: `Unexpected spaces found. [warning]`

ironsidevsquincy: That's a editor config error i think (rather than eslint). Make sure your node modules are up to date

ironsidevsquincy: Actually, think you'll need to update `lintspaces-cli` to the latest version (`0.4.0`)

test/server/transforms/special-reports-hub-featured.test.js
```diff
@@ -1,3 +1,5 @@
+/*eslint no-console: 0*/
```
ironsidevsquincy: does this test file need a console log?

PR #1399, opened 2016-09-22
a11y

lc512k:
a11y checks in CI. Errors won't break the build (yet)

Only ~3 errors per page. Well done!

screen shot 2016-09-22 at 11 08 40

cc @andygnewman @rowanmanning


.pa11yci.js
```diff
@@ -0,0 +1,18 @@
...
+				"height": 480
+			}
+		}
+	},
+	"urls": [
+		process.env.TEST_URL + "news-feed",
+		process.env.TEST_URL + "special-reports",
+		process.env.TEST_URL + "world",
+		process.env.TEST_URL + "stream/regionsId/TnN0ZWluX0dMX0VT-R0w%3D",
+
```
andygnewman: Might be good to put in;
life and arts hub page (`life-arts`)
fast FT (`fastft`)

As these are very different designs and you could probably just use special-reports and world from the existing list as the other pages are using components already featured on those pages.

lc512k: Thanks @andygnewman, will add those in 👍

PR #1401, opened 2016-09-23
Top stories on all streams

andygnewman:
This enables any stream to be able to pick up a top stories (and opinion & analysis) list when published by editorial with no further code changes.

Streams excepted from this are;
- FastFT
- news feed
- Photo Diary
- Podcast Hub & Podcast Shows


server/config/profiles.js
```diff
@@ -26,14 +26,20 @@ const lightSignup = require('../transforms/light-signup');
 const jobsBox = require('../transforms/jobs-box');
 const specialReportsHubFeatured = require('../transforms/special-reports-hub-featured');
 
-const defaultTransforms = [
+const defaultTransformsBasic = [
```
bjfletcher: interesting re using adjective "basic", guess you find this makes better sense than the noun "base"?

bjfletcher:
Very nice. Looking forward to seeing the fruits of this. Do you know what URLs to check once this gets shipped? Perhaps there's some URL telling us which ones have been curated?

andygnewman:
Unfortunately there's no way of understanding which have been curated.
Wanted to do this more to test system impact rather than pick up any new lists.

It will increase our CAPI hits from stream pages, but should be OK.

PR #1402, opened 2016-09-26
Ads 1047

andrewgeorgiou1981:
@adgad @VladDubrovskis @gvonkoss


views/article-list.html
```diff
@@ -93,6 +95,26 @@
 				{{#if hasStreamListTitle}} data-has-stream-list-title{{/if}}
 			>
 				<div class="o-grid-row">
+					{{#if ads.twoMidRailAds}}
+
+						{{#unless properties.streamListTrendingTopics}}
+							{{#with trendingTopics}}
```
adgad: How does this work? If you're in an unless streamListTrendingTopics wouldn't that mean there would be no trendingTopics property?

views/article-list.html
```diff
@@ -93,6 +95,26 @@
...
+									{{{reactRenderToString ../components.Section spread=@this}}}
+								</div>
+							{{/with}}
+						{{/unless}}
+
+						{{#if topSection}}
+							{{>ads/mid1-both-large}}
+						{{else}}
+							{{>ads/mid-both-large}}
+							<br><br><br>
```
adgad: What's this for?

views/partials/ads/mid1-both-large.html
```diff
@@ -0,0 +1,14 @@
+<div
+	class="o-ads advert--mid1-both-large"
+	data-o-ads-name="mid1-both-large"
+	data-o-ads-targeting="pos=mid1;"
+	data-o-ads-center="true"
+	data-o-ads-label="false"
+	data-o-ads-formats-default="MediumRectangle,HalfPage,Responsive"
```
adgad: I'm pretty sure default isn't a thing anymore, maybe we could delete it from these templates as it looks a little odd?

views/article-list.html
andrewgeorgiou1981: removed this.

views/article-list.html
andrewgeorgiou1981: Correct - this was a bug with my code, i've corrected it. good spot.

PR #1404, opened 2016-09-27
Tour tip

laurieboyes:
Get the tour tip into the streams page. After the 6th article on desktop and the 18th article on mobile (to make room for ads)

There's a bit of style tweaking to get the tip to line up nicely with the articles (more on this to follow).

Note that the images will be updated soon to up the contrast between them and the page background.

Desktop:
screen shot 2016-09-27 at 16 51 16

Mobile:
screen shot 2016-09-27 at 16 51 01


client/components/tour-tip/_main.scss
```diff
@@ -0,0 +1,53 @@
...
+			position: absolute;
+			height: 1px;
+			left: 0;
+			right: 0;
+			top: -1px;
+			background-color: #cec6b9;
+		}
+	}
+
+	// some heinous overriding to line it up with articles in streams
```
laurieboyes: I'm of the opinion that these overrides here are unacceptable. Potential improvements and alternatives:

- New class in n-ui `.tour-tip--stream-page` which performs these overrides but as least keeps it all in one place (also need some way of setting this class)
- Leave it as it is (it looks nasty) 
- ???

Without this stuff:
screen shot 2016-09-27 at 17 00 29

client/components/tour-tip/_main.scss
```diff
@@ -0,0 +1,53 @@
...
+	@include oGridRespondTo(L) {
+		display: block;
+	}
+}
+
+.tour-tip-stream-container {
+
+	margin-top: $spacing-unit * 2;
+
+	.tour-tip {
```
keirog: Couldn't we just use border here?

client/components/tour-tip/_main.scss
```diff
@@ -0,0 +1,53 @@
...
+		margin: 10px;
+	}
+
+	.tour-tip__standout {
+		line-height: 31px;
+		font-size: 26px;
+	}
+
+	.tour-tip__img-container {
+		padding: 0;
```
keirog: _Shouldn’t_ need this bit now

client/n-ui-config.js
```diff
@@ -1,6 +1,7 @@
 module.exports = {
 	preset: 'complete',
 	features: {
-		lazyLoadImages: true
+		lazyLoadImages: true,
+		tourTip: true
```
keirog: Is this needed?

client/components/tour-tip/_main.scss
```diff
@@ -0,0 +1,49 @@
...
+		padding-left:0;
+		padding-right: 0;
+	}
+
+	.tour-tip__text {
+		margin: 10px;
+	}
+
+	.tour-tip__standout {
+		line-height: 31px;
```
keirog: Please use the [o-typography mixin](https://github.com/Financial-Times/o-typography#serif) for this (where size `l` of `Serif` outputs these exact values, so should be a straight swap)

client/components/tour-tip/_main.scss
```diff
@@ -0,0 +1,49 @@
...
+
+	// some heinous overriding to line it up with articles in streams
+
+	.tour-tip__main {
+		padding-left:0;
+		padding-right: 0;
+	}
+
+	.tour-tip__text {
+		margin: 10px;
```
keirog: Not sure about this being different from the other tips

client/components/tour-tip/_main.scss
laurieboyes: We managed to clean up and reduce the number of overrides to the point that we're willing to accept it

PR #1411, opened 2016-09-29
Dedupe special reports

andygout:
cc @ironsidevsquincy @bjfletcher 

On [Special Reports page](https://www.ft.com/special-reports) will prevent duplication of reports included in Featured from appearing in Latest.


server/lib/special-report-mapping-poller.js
```diff
@@ -0,0 +1,29 @@
+/* eslint-disable new-cap */
```
ironsidevsquincy: ?

server/lib/special-report-mapping-poller.js
```diff
@@ -0,0 +1,29 @@
...
+
+import Poller from 'ft-poller';
+
+const parseData = data => data.filter(row => row.interactiveurl.includes('https://www.ft.com/special-reports/'));
```
ironsidevsquincy: `includes` or `startsWith`? aka `indexOf(...) === 0`

server/controllers/special-reports.js
```diff
@@ -6,6 +6,23 @@ const longTitle = require('../lib/long-title');
 const logger = require('@financial-times/n-logger').default;
 const profiles = require('../config/profiles');
 const TransformSeries = require('../lib/transform-series');
+const specialReportMappingPoller = require('../lib/special-report-mapping-poller');
+
+const dedupeSpecialReports = res => {
```
ironsidevsquincy: rather than taking a`res` object and having to drill down, make this function more localised and just accept the arguments it's going to use, e.g.

```
const dedupeSpecialReports = (topSpecialReports, allSpecialReports => {
```

Where 
 * `topSpecialReports = res.templateData.topSection.leadStory.data.content.concat(res.templateData.topSection. columnOne.data.content)`
 * `allSpecialReports = res.templateData.stream.items`

server/lib/special-report-mapping-poller.js
andygout: Whoops.

server/lib/special-report-mapping-poller.js
andygout: Is there anything against using `includes` (maybe a performance issue)? Took it from [here](http://stackoverflow.com/questions/1789945/how-to-check-if-one-string-contains-another-substring-in-javascript#answer-19486875).

server/controllers/special-reports.js
andygout: 👍

server/controllers/special-reports.js
```diff
@@ -53,7 +68,14 @@ module.exports = (req, res, next) => {
 		.then(() => transform.post(res, flags))
 
 		// Finally render the special reports hub list
-		.then(() => renderStream(req, res)(res.templateData))
+		.then(() => {
+			res.templateData.stream.items = dedupeSpecialReports(
+					res.templateData.topSection.leadStory.data.content.concat(res.templateData.topSection.columnOne.data.content),
```
andygout: This line is causing tests to fail: `"Cannot read property 'leadStory' of undefined"`. Seems to be an issue with asynchronicity in that this property should be available once `transform.post` has finished running. Am looking into it.

server/lib/special-report-mapping-poller.js
ironsidevsquincy: Not at all, but in this case you're looking for interactiveUrls that _start with_ that url snippet, not _include_ it, e.g. this would match `http://foo.com?url=https://www.ft.com/special-reports/` which you don't want

server/controllers/special-reports.js
ironsidevsquincy: I'm guessing there isn't any `topSection` data in the test fixture

PR #1414, opened 2016-09-30
Fix left column being off screen on IE in L breakpoint

bjfletcher:
On IE, despite the `display: block`, `flex-basis` still has an effect.


ironsidevsquincy:
👍

PR #1423, opened 2016-10-03
Add: Most Read component to Special Reports hub page

andygout:
cc @bjfletcher 

screen shot 2016-10-03 at 15 54 50


server/controllers/special-reports.js
```diff
@@ -27,6 +27,8 @@ module.exports = (req, res, next) => {
 	res.set('Surrogate-Control', res.FT_HOUR_CACHE);
 	res.set('Cache-Control', res.FT_NO_CACHE);
 
+	req.params = Object.assign({}, req.params, { thingId: 'MTA=-R2VucmVz' });
```
andygout: This feels wrong, but given Special Reports is such a special case, not entirely sure how else to add this `thingId` value used by GraphQL.

bjfletcher:
Looks good 
\o/

views/special-reports.html
```diff
@@ -14,6 +14,15 @@
 		{{#if stream.items}}
 			{{#if topSection}}
 				{{>top-section-default}}
+
+				{{#ifAll mostRead flags.streamSpecialReportsMostRead}}
+					{{#with mostRead}}
```
ironsidevsquincy: any need for this if you're checking it's existence above? just pass it straight in to `reactRenderToString`?

server/controllers/special-reports.js
ironsidevsquincy: any reason not to add this straight to params?

`req.params. thingId = 'MTA=-R2VucmVz';`

ironsidevsquincy: also, how is this used by GraphQL?

andygout: Sets `conceptId` as `concept` value in [`server/transforms/most-read`](https://github.com/Financial-Times/next-stream-page/blob/52cc7d7f488d8bc6f0f730f13883b8f8c5525493/server/transforms/most-read.js#L16-L17) which is then used in [`config/graphql-queries`](https://github.com/Financial-Times/next-stream-page/blob/c187fd109db0e91d804fe0d0af23dd9c111b635c/config/graphql-queries.js#L81).

views/special-reports.html
andygout: For [other instances](https://github.com/search?utf8=%E2%9C%93&q=user%3AFinancial-Times+reactRenderToString&type=Code&ref=searchresults) of `reactRenderToString` we pass in `spread=@this`; can't see how you pass in an object directly (am looking through the docs).

ironsidevsquincy: `@this` is just a reference to the current scope, which you're setting with the `with` to `mostRead` (it also only runs the contents if the variable isn't empty

views/special-reports.html
```diff
@@ -14,6 +14,13 @@
 		{{#if stream.items}}
 			{{#if topSection}}
 				{{>top-section-default}}
+
+				{{#ifAll mostRead flags.streamSpecialReportsMostRead}}
+					<div id="{{id}}"{{#if trackScrollEvent}} class="js-track-scroll-event"{{/if}}>
```
ironsidevsquincy: does this need to be `{{mostRead.id}}` now?

ironsidevsquincy: and `mostRead. trackScrollEvent`?

PR #1427, opened 2016-10-03
Remove: Handlebars '#with' for article list Most Read component

andygout:
cc @bjfletcher 

Something I learned working on this branch: https://github.com/Financial-Times/next-stream-page/pull/1423


ironsidevsquincy:
This one you could probably leave the `with` and get rid of the `ifAll`, right? As `mostRead` will never exist if the flag is off - https://github.com/Financial-Times/next-stream-page/blob/52cc7d7f488d8bc6f0f730f13883b8f8c5525493/server/transforms/most-read.js#L14

PR #1430, opened 2016-10-04
Add: Assign thingId transform for use by Special Reports

andygout:
cc @ironsidevsquincy  @bjfletcher


server/config/profiles.js
```diff
@@ -1,6 +1,7 @@
 const podcasts = require('n-podcast-mapping').primaryTags();
 
 const articleModel = require('../transforms/article-model');
+const assignThingId = require('../transforms/assign-thing-id');
```
ironsidevsquincy: is this the best name for this transform? how about what it's doing, e.g. `addXMostRead` (whatever most read it is)

server/transforms/assign-thing-id.js
```diff
@@ -0,0 +1,7 @@
+module.exports = () => {
+	return class {
+		pre (req) {
+			if (req.url === '/special-reports') req.params.thingId = 'MTA=-R2VucmVz';
```
ironsidevsquincy: won't the url always have that, as the profile above is already doing this test

server/config/profiles.js
ironsidevsquincy: Or maybe just `assign-X-to-thing-id`

What's is the `thingId` anyway?

server/transforms/assign-thing-id.js
andygout: True, and I suppose in the instance of another profile using this transform we'd only then need to worry about customising for respective profiles.

server/config/profiles.js
andygout: Breaking it down, it's `genreId` used by its non-vanity URL (`/stream/genreId/MTA=-R2VucmVz`), which (in this context) is assigned to the conceptId for use in the GraphQL query to acquire its Most Read articles.

On that basis do we want something like `assign-graphql-id-value-to-thing-id` / `assign-genre-id-value-to-thing-id` / `assign-url-value-to-thing-id`?

PR #1433, opened 2016-10-04
Add: Additional article in top section

andygout:
cc @bjfletcher 

### Top Stories Column Two
top-stories-column-two

### Opinion & Analysis from API
opinion-analysis-from-api

### Opinion & Analysis from Stream List
opinion-analysis-from-stream-list


server/transforms/top-section.js
```diff
@@ -67,7 +67,7 @@ module.exports = () => {
 								topSection.columnOne = getSection('top-stories-column-one', {content: topStories.slice(1,5)}, flags);
 							} else {
 								topStories.shift();
-								const splitPoint = Math.ceil(topStories.length/2);
+								const splitPoint = Math.floor(topStories.length/2);
```
bjfletcher: wouldn't this mean, if there are 7 cards, 3 on left, 4 on right?

test/server/lib/opinion-and-analysis.test.js
```diff
@@ -7,8 +7,8 @@ const concordanceStub = sinon.stub();
 const listsStub = sinon.stub();
 
 const fixtureList = require('../../fixtures/capi-list-opinion-analysis');
-const fixtureListThree = JSON.parse(JSON.stringify(fixtureList));
-fixtureListThree.items.splice(3, fixtureListThree.items.length - 3);
+const fixtureListFour = JSON.parse(JSON.stringify(fixtureList));
+fixtureListFour.items.splice(4, fixtureListFour.items.length - 4);
```
bjfletcher: no problem using splice as this is pattern adopted elsewhere, but in the future, it'd be good to use the more immutable approach - slice instead of splice... mutating state tends to lead to hard-to-find bugs

server/transforms/top-section.js
andygout: Yup - that is new requirement.

test/server/lib/opinion-and-analysis.test.js
andygout: Think this is worth fixing now; will add another commit shortly.

bjfletcher:
Very, very nice \o/ 🎉

PR #1440, opened 2016-10-06
fix: A bug in the stream page causing 4th biggest 5xx error on Next in the past 24h

matthew-andrews:
`.reduce` needs an `initialValue`


server/controllers/related/tag.js
```diff
@@ -19,7 +19,7 @@ module.exports = function (req, res) {
 				)
 				// flatten
 				.reduce(
-					(prev, relations) => prev.concat(relations)
+					(prev, rels) => prev.concat(rels), []
```
matthew-andrews: renamed this to make it clear it's a different array

server/controllers/related/tag.js
```diff
@@ -47,7 +47,8 @@ module.exports = function (req, res) {
 				relations: relations
 			});
 		})
-		.catch(
-			() => res.status(500).send()
-		);
+		.catch(err => {
+			logger.error({ event: 'RELATED_SPECIAL_REPORTS_ERROR', err });
```
matthew-andrews: added a logger so that we can see this in the logs

bjfletcher:
```
> let t = [1, 2, 3].map(t => t+1);
undefined
> t
[2, 3, 4]
```

madness

this is a great example of why immutability/FP school of thought helps us a lot around hard-to-find bugs

we should also perhaps do an audit of all our codebase, replacing `let` where it is meant to be `const`

matthew-andrews:
do you agree with my changes @bjfletcher ?

matthew-andrews:
your code snippet makes sense to me ……

bjfletcher:
@matthew-andrews yes, most absolutely... apart from the logging change suggested by Darren

server/controllers/related/tag.js
matthew-andrews: oo

PR #1441, opened 2016-10-06
Turn off autofocus for stream filters input

i-like-robots:
https://github.com/i-like-robots/react-tags#autofocusOption


i-like-robots:
The input component should also be updated to 5.x

i-like-robots:
wrong place

PR #1443, opened 2016-10-06
Set autofocus to false for react tags

i-like-robots:
https://github.com/i-like-robots/react-tags#autofocusOption


i-like-robots:
Bundled with #1444

PR #1444, opened 2016-10-06
Update react tags

i-like-robots:
Update the react tag autocomplete component to v5 (req. a change of CSS class names, otherwise all the same), https://github.com/i-like-robots/react-tags/blob/master/README.md#upgrading-from-4x-to-50

Also bumps node-sass to 3.10.x to shut up all the C internals on Node 6.x

Also sets `autofocus=false`


matthew-andrews:
🚢

andygout:
Late to the party but 👍

PR #1447, opened 2016-10-07
Add: Top margin to top section if custom fragment and meta title present

andygout:
cc @bjfletcher 

Adds some breathing space below the custom fragment if a meta title is being used.

#### From
from

#### To
to


shared/components/custom-fragment/_main.scss
```diff
@@ -17,3 +17,9 @@
 		margin-top: 20px;
 	}
 }
+
+.custom-fragment + .stream-section__top {
+	.section__column--meta {
+		margin-top: 20px;
+	}
+}
```
ironsidevsquincy: personal preference, but I keep the element that's actually being styled at the 'top level', makes it easier than having to dig down into a selector, e.g.

```
.section__column--meta {
    .custom-fragment + .stream-section__top & {
        margin-top: 20px;
    }
}
```

andygout: Yup - that does make more sense - will change now.

bjfletcher:
I think the new agreement is that we remove the top meta title. So, just hide it or remove it altogether. That is, the "Top Stories" can be gone. A lot like how it is on frontpage.

andygout:
Oh, thought that was just for Special Reports. Will draw up a card for that work.

bjfletcher:
👍 looks like what we want won't happen for a while, so this is definitely my 2nd preference. Let's get this in :)

PR #1448, opened 2016-10-10
fix: Give news-feed a canonical URL

matthew-andrews:


bjfletcher:
:+1: noticed it was undefined on news-feed... so let's have this in

PR #1449, opened 2016-10-10
Add tour-tip styling to live stream pages

keirog:
Before:
before

After:
after


debugwand:
Does this affect other stream pages?

keirog:
It shouldn’t have any visual effect, just adds a bit of CSS

PR #1452, opened 2016-10-12
Ads right rail padding fix

andrewgeorgiou1981:
@adgad @VladDubrovskis @gvonkoss


andygnewman:
We added a 'properties' property onto the data that is used in the template for streams.

https://github.com/Financial-Times/next-stream-page/blob/master/server/lib/presentation/properties.js

Would be good if you could use those properties in the template conditionals.

Example https://github.com/Financial-Times/next-stream-page/blob/master/views/article-list.html#L103

PR #1455, opened 2016-10-13
Insert custom fragment before item

kavanagh:
Custom fragments are currently inserted after a specific item in the stream. This changes to that to insert the fragment before an item.


bjfletcher:
I'm on the case... keep this PR open please.

PR #1456, opened 2016-10-13
Remove references to next-card

adgad:
@andygnewman @andygout @keirog @GlynnPhillips 
Doesn't look like this was actively used anywhere anyway, so remove it all.


andygnewman:
Think there might be some CSS that is imported from next-card

adgad:
Yes this is a bigger change than I thought 😞

adgad:
@andygnewman This okay now?

andygnewman:
LGTM 👍

PR #1458, opened 2016-10-14
in readiness of a new page with clearer templating system and using n-teaser

bjfletcher:
- reduce duplication across article-list and hub pages
- this new page needs just the light signup, so creating new "minimum"
  transforms list

note: n-teaser is n-card's replacement
https://github.com/financial-times/n-teaser


andygnewman:
Nice removal of duplication in the templates.

Not sure I understand the use case for moving lightSignUp transform into a default category on it's own?

bjfletcher:
@andygnewman all will be clearer in the next PR...

PR #1459, opened 2016-10-14
Amend: Bring special reports pagination data in line with others

andygnewman:
cc: @bjfletcher 

On special reports page, the pagination data object was at a different level in the template data model than other stream pages (so the isPage property was not available).

This sorts that out.


server/controllers/special-reports.js
```diff
@@ -58,11 +58,11 @@ module.exports = (req, res, next) => {
 								longTitle: longTitle(report.key),
 								publishedDate: report.mostRecent.date.value_as_string
 							};
-						})
+						}),
+					pagination: req.perPage.toObject()
 					},
```
bjfletcher: the indentation doesn't look right here

PR #1460, opened 2016-10-14
Ads right rail padding fix

andrewgeorgiou1981:
@VladDubrovskis @adgad @gvonkoss @andygnewman


views/article-list.html
```diff
@@ -77,7 +77,16 @@
...
 					{{#if ads.streamListSecondMidAd}}
-						{{>ads/mid-sidebar-large pos="mid1" nonResponsiveOnly=ads.fullWidthResponsive}}
+						{{#unless properties.hasMostRead}}
+							{{#if properties.hasTopSection}}
+								{{>ads/mid-sidebar-large pos="mid1" extraPadding="true" nonResponsiveOnly=ads.fullWidthResponsive}}
+							{{else}}
+								{{>ads/mid-sidebar-large pos="mid1" nonResponsiveOnly=ads.fullWidthResponsive}}
+							{{/if}}
+						{{/unless}}
+						{{#with properties.hasMostRead}}
```
adgad: Should this be #if instead of #with (i think they both probably work, but if makes a bit more sense here.

client/components/ads/_main.scss
```diff
@@ -22,6 +22,10 @@ $ad-medium-breakpoint: 760px;
 	margin-top: 100vh;
 }
 
+.advert--extra-padding {
```
adgad: What's the difference between this and the one above? Can this be margin, and then can the two use cases above just add this class?

views/article-list.html
andrewgeorgiou1981: yep, updated.

client/components/ads/_main.scss
andrewgeorgiou1981: done.

PR #1462, opened 2016-10-14
A new page

bjfletcher:
This PR kills two birds with one stone:

1. a new page
2. explore n-section/React-free ways of doing things, and start using n-teaser

image

As you can see, much of what's in the new `.css` should probably be in `n-teaser` or `n-teaser-collection`

The very heavily nested structure in the new `.html` is something the templating team will want to work on in the coming weeks.

(Note that head files are removed from `nodemon` because they're now
automatically reloaded by `n-express`.)


server/config/profiles.js
```diff
@@ -177,6 +178,16 @@ module.exports = [
 	},
 
 	{
+		name: 'Future of Britain',
+		match: (req) => req.params.thingId === 'MWUzYzljMTEtMmU1NS00OTMzLWJiMGEtNDQ5NzgyYmJiNDBh-VG9waWNz',
+		transforms: [
+			customFragments(),
```
andygnewman: why not put customFragments into the defaultTransformsMinimum?

server/transforms/future-britain.js
```diff
@@ -0,0 +1,66 @@
...
+			tag: {
+				href: item.brand.url,
+				text: item.brand.name
+			},
+		} : {
+			img: {
+				alt: item.mainImage.description,
+				src: item.mainImage.url
+			},
+			tag: {
```
andygnewman: tag is returned by the decorator, so not sure why you're duplicating logic?
Plus the metadata is in no specific order, so doing metadata[0] is wrong.
n-content-decorator assigns the best one to the tag property - which should be used if there is no brand.

server/transforms/future-britain.js
```diff
@@ -0,0 +1,66 @@
...
+// this is the same as contentDecorator except that it's using n-tosser's model
+const convert = items => {
+	return items
+		.map(item => {
+			contentDecorator(item, { mutateOriginal: true });
+			return item;
+		})
+		.map(item => Object.assign({
+			heading: {
+				href: item.url,
```
andygnewman: not sure why you're renaming stuff? can't you just use as it is? we get in to lots of difficulty with the overall data if we keep on calling it different things at different stages. Should ideally stick with what it was originally named as.

bjfletcher: @andygnewman all this model work is interim, because I know that you're working on a replacement model.  :) Will use tag instead of `mainImage` (reason was that `tag` didn't have the image description for the `alt` attribute)

PR #1463, opened 2016-10-18
add the sticky header style and put it as first thing after body element

VladDubrovskis:
/cc @adgad @andrewgeorgiou1981 @gvonkoss


andrewgeorgiou1981:
👍

PR #1468, opened 2016-10-20
[templating] Case study: dumb component used only server-side

bjfletcher:
This demonstrates the change to Handlebars from React for dumb React components that are server-side only.

If everyone's happy, will do the same to other, similar components.


andygnewman:
I'm happy with this approach.

PR #1472, opened 2016-10-27
fix the breaking page by correctly importing the n-logger

VladDubrovskis:
/cc @andygnewman @bjfletcher 

After investigating what have broken the Recruitment page before, it seems the reason was the way we import the n-logger.

Previous way was resulting in undefined.

Might need to be updated in other places too? Just wanted a pair of eyes to have a look and verify that is the case


VladDubrovskis:
Think the reason behind this is that the installed version is not default JS but rather transpiled ES6 via Babel?

ironsidevsquincy:
Yep, that's the correct way to import it, it's the default export

https://github.com/Financial-Times/n-logger/blob/master/src/main.js#L96

VladDubrovskis:
In codebase there are different ways used to import it, e.g.:
https://github.com/Financial-Times/next-stream-page/blob/ae2cb07cb3a6a8b12cd257de3706f193d9cb518e/server/transforms/custom-fragments.js#L2
vs 
https://github.com/Financial-Times/next-stream-page/blob/6384d7b949e9d660a341f3f97153cab0a892f92c/server/lib/custom-fragments-poller.js#L6

Will update the other `import { logger}` to `import logger` as well

ironsidevsquincy:
Oh, defo

andygnewman:
Thanks @VladDubrovskis

PR #1473, opened 2016-10-27
Switch next-graphql-api requests over to next-api

ironsidevsquincy:


ironsidevsquincy:
Really need to write a linting rule to catch `only`s

Could you give this a plus one if you're happy

PR #1475, opened 2016-11-01
Fix pagination bug

debugwand:
Stops the user trying to search for results over result number 4000 as this will just result in a 422 error from the API https://developer.ft.com/docs/api_v1_reference/search/tutorial#Paging


GlynnPhillips:
LGTM

PR #1483, opened 2016-11-04
Use red and black fastft logo

JakeChampion:
After talking with Max Mockett and editorial, we've agreed that the logo should be red and black for fastFT.


JakeChampion:
screen shot 2016-11-04 at 18 23 39

PR #1498, opened 2016-11-17
started work on adding surrogate keys for concept ids and lists

wheresrhys:
@phamann @andygnewman 

This
- adds surrogate key headers for stream id and list ids
- removes dependency on the preditor pages api
- memoizes concurrency api responses (I've checked with Sarah Wells that it's ok to do so)


wheresrhys:
Blocked by the lack of a capi2 list for /world/us top stories

wheresrhys:
unblocked thanks to Mus

PR #1499, opened 2016-11-17
Fix: unique ids on columnist elements

andygnewman:
@lc512k 

This fixes the duplicate Columnist ids


lc512k:
Thanks, @andygnewman

PR #1503, opened 2016-11-18
Ai curate page

bjfletcher:
This is plumbing just to enable us to have a point in to which we can curate the top stories & opinion/analysis lists using an automatic method.

This is all behind the `aiCuratePage` flag and the automatic method for curating the lists is hardcoded and very simple and definitely temporary (note how it does ES search :) - we'll replace it - it's just temporary) as follows:

a) top stories - the most recent 5 stream items that aren't `'Comment', 'Special Report', 'Analysis'`
b) opinion/analysis - the most recent 5 stream items that are `'Comment', 'Special Report', 'Analysis'`

In the coming weeks, we'll be replacing these with better models based on more factors.

Note that I've extracted "decoration" bits from the `server/lib/top-stories.js` module into a separate module so that it can be reused.


server/lib/ai-curated-stories.js
```diff
@@ -0,0 +1,45 @@
...
+
+module.exports = (streamId) => {
+	// here is where we curate the top stories & analysis/opinion
+	// at the moment, it's just the most recent ones
+	return fetchItems(streamId)
+	.then(items => {
+		const topStoryIds = getTopStories(items)
+			.slice(0, 5)
+			.map(item => item.id);
+		return decorateTopStories(streamId, Promise.resolve(topStoryIds))
```
andygnewman: Does topStoryIds need to be in a promise resolve?

server/lib/ai-curated-stories.js
```diff
@@ -0,0 +1,45 @@
...
+		.then(topStories => {
+			const opinionAnalysisIds = getOpinionAnalysis(items)
+				.map(item => item.id)
+				.filter(item => topStoryIds.indexOf(item) === -1)
```
andygnewman: I guess you're future proofing, but the way you've calculated Top Stories vs Opinion and Analysis should not have any dupes.

server/lib/ai-curated-stories.js
```diff
@@ -0,0 +1,45 @@
...
+				.slice(0, 5);
+			return api.content({ uuid: opinionAnalysisIds })
```
andygnewman: If you take the content from the res object, it should already be fully populated and you wouldn't need to run this additional api call.

server/lib/ai-curated-stories.js
bjfletcher: Ha! Well spotted - very sharp you are now. :)

PR #1505, opened 2016-11-21
Move teaser stream heading into the collection template

keirog:
Also removes the redundant `stream__container`.

@adgad


keirog:
(Easier to read [with `w=1` on](https://github.com/Financial-Times/next-stream-page/pull/1505/files?w=1), or individual commits)

PR #1507, opened 2016-11-21
Cut number of trending topics down to 2

keirog:
@adgad


andygnewman:
😿

keirog:
😞  A merge too far? Sorry to upset, please elaborate

andygnewman:
Just liked having three, that's all. The redo of streams seems to not be data driven in anyway, just an arbitrary redesign based on what might look nice, more so than what works for the reader. (Not that I necessarily think that the third trending topic got many clicks, but did anyone check?)

PR #1528, opened 2016-11-23
Seasonal appeal

bjfletcher:
Added a profile for Seasonal appeal page.

Abstracted explainer so that this can be used also.

image


andygnewman:
I thought that the ad would move up to the right hand side of the top story?

bjfletcher:
@andygnewman The page is split into two sections, as you know, the top section and the bottom section (with stream list).

The top section's right hand side has the very tall ad, too tall that it causes whitespace between the lead story and the bottom section.

So I hide that in `AdProposition`.  Allowing bottom section's right hand ad to show.

It seems to me that `AdProposition` doesn't allow us to change the double ad to single ad for top section's right hand.

andygnewman:
Makes sense - I had a conversation with @mmockett when he was coming up with this simpler idea, and we were thinking that ads could decide to only serve a single MPU into that spot, so we could not have the whitespace up there. Have asked Max to clarify whether he has an answer for that.

If ad's can't configure it their side, we could in the ad markup only allow for a single MPU for that stream page (but that feels like a lot of bespoking for this one instance.

server/transforms/explainer.js
```diff
@@ -0,0 +1,44 @@
+const fetchExplainerData = (req) => {
+	// FIXME we should perhaps plumb this data in from somewhere that editors edit
```
andygnewman: Bertha? Kidding, for now - but if this is a longer term solution to editorial customising stream pages then perhaps a Bertha sheet is the way to go.
Can wait until that comes up in 2017.

server/transforms/explainer.js
```diff
@@ -0,0 +1,44 @@
...
+			title: 'About Special Reports',
+			paragraph1: 'Special Reports provide in-depth FT coverage of countries around the world, as well as industries from tech to luxury, and themes ranging from workplace health to entrepreneurship.',
+			paragraph2: 'They appear in a variety of formats, from magazines and broadsheets, to video and interactive stories.',
+			isSpecialReports: true
+		};
+	} else if (req.params.thingId === 'ZDQyNmY4NDEtZTUxMS00NzcxLTk2M2MtYWI5ZTQzMzI2NmNl-VG9waWNz') {
+		// Seasonal Appeal page
+		data = {
+			imgSrc: 'https://www.ft.com/__origami/service/image/v2/images/raw/https%3A%2F%2Fnext-geebee.ft.com%2Fassets%2Fspecial-reports-hero.jpg?source=ft&width=600',
+			imgAlt: 'A collage of special reports',
```
andygnewman: Too much copy & paste...

mmockett:
@adgad - do you know if Ads can target an MPU on this specific stream page? The slot is common across all Streams with Top Sections, but because we're only serving a single story into that in this instance, a DMPU wouldn't work.

server/transforms/explainer.js
```diff
@@ -0,0 +1,44 @@
...
+			title: 'Doctors on the Frontline',
+			paragraph1: 'In the world\'s most troubled places, where conflict, disaster and disease brings terrible suffering, Médecins Sans Frontières gives free, independent medical care to people in desperate need.',
+			paragraph2: 'Through the FT\'s Seasonal Appeal you can support this pioneering humanitarian organisation. FT writers have travelled with MSF across four continents - from Iraq to Haiti - to witness its life-saving work. You will find their extraordinary stories here, told in words, photographs and film.',
+			buttonLabel: 'Donate',
+			buttonLink: 'https://msf.org.uk/ft'
+
+		}
+	}
+	return Promise.resolve(data);
```
andygnewman: not sure why you're doing this as a promise.

You could just evaluate the `isFirstPage` on the pre and then just assign the correct data to res.templateData.explainer on the res.

bjfletcher: Yes, I like the flexibility it gives us in that we can do some API call instead when we come to it.

views/partials/explainer.html
```diff
@@ -1,19 +1,26 @@
...
 		</div>
 		<div class="explainer__clear"></div>
 	</div>
-	<div class="explainer__actions">
-		<a href="http://fttoolkit.co.uk/d/calendar/" class="explainer__link" data-trackable="plans">Future and planned reports</a>
-		<a href="http://www.ft.com/reports-library?ft_site=falcon&desktop=true" class="explainer__link" data-trackable="library">Reports library</a>
-		<div class="explainer__follow">
-			<span class="explainer__follow-note">Follow special reports</span>
-			<a class="explainer__twitter o-share__action o-share__action--twitter" target="_blank" href="http://twitter.com/ftreports" data-trackable="twitter"><i>Twitter</i></a>
+	{{#if isSpecialReports}}
```
andygnewman: Whilst you've done a fair bit of abstracting, this looks a little at odds with the rest of the set-up.

With only two options it's probably OK, but definitely a candidate for refactoring if we use it more.

PR #1530, opened 2016-11-23
Remove permalink hack

commuterjoy:
I think was responsible for 13% of router timeout errors this week. Either the WordPress API it slow, or there's some redirection logic chain that's holding things up. Removing it to see if it improves matters.


commuterjoy:
https://github.com/Financial-Times/next-redirector/pull/128

wheresrhys:
:+1:

PR #1538, opened 2016-11-29
fix pagination issue on special reports and add pagination smoke tests

bjfletcher:


test/smoke.js
```diff
@@ -9,9 +9,11 @@ module.exports = [{
 		'/stream/specialReportsId/MDEzZjdiOTAtYTFhNC00NzBjLTnotareport': 404,
 		//Section pages
 		'/stream/sectionsId/MQ==-U2VjdGlvbnM=': 200, // standard section page '/world'
+		'/stream/sectionsId/MQ==-U2VjdGlvbnM=?page=2': 200, // pagination on '/world'
 		'/stream/sectionsId/MTU4-U2VjdGlvbnM=': 200, // lifestyle layout section page '/life-arts/style'
 		//Hub pages
 		'/stream/sectionsId/MTQ4-U2VjdGlvbnM=': 200, // page constructed purely from lists '/life-arts'
+		'/stream/sectionsId/MTU4-U2VjdGlvbnM=': 200, // style page
```
andygnewman: style page isn't a hub page - is a normal section/stream construct, just uses the top stories style of the hub page.

andygnewman: So would put in the section above.

PR #1539, opened 2016-11-29
Fix: data-trackables into top sections with teasers

andygnewman:
cc: @adambraimbridge 

puts back in the data trackables as per existing.


adambraimbridge:
Excellent 👍

PR #1540, opened 2016-11-29
Amend: Swap from api to coco

andygnewman:
Need to check with UPP team before merging in.

API still used as backup for concordance as experience tells us that coco hasn't always had the same volume of concordance as the api entry point.


andygnewman:
Happened with switch to next-api anyway.

PR #1549, opened 2016-12-05
Watch for Handlebars template file changes

georgecrawford:


georgecrawford:
cc @rowanbeentje @ironsidevsquincy

nodemon.json
```diff
@@ -2,6 +2,8 @@
   "ignoreRoot": [],
   "watch": [
     "node_modules",
-    "server"
-  ]
+    "server",
+    "views"
+  ],
+  "ext": "js json html"
```
ironsidevsquincy: Personally wouldn't bother with this as there's not really anything else in those dirs, but if you do you might want to add `css` (our head.css file is written inline, so needs an app restart if it changes) - https://github.com/Financial-Times/next-front-page/blob/master/nodemon.json#L6

PR #1552, opened 2016-12-05
Introduce storylines

TheoLeanse:
@gvonkoss @andygnewman


andygnewman:
LGTM 👍

server/lib/storyline-data.js
```diff
@@ -0,0 +1,10 @@
+const fetchres = require('fetchres');
+const STORYLINES_API_URL = 'https://next-storylines-api.herokuapp.com';
```
andygnewman: Suggest move this into a config vars set up

PR #1554, opened 2016-12-08
AI curation

bjfletcher:
Some plumbing so that we can use Amazon Machine Learning models to curate stream pages.


missamynicholson:
👀 looks good, Ben!

PR #1560, opened 2016-12-13
Adds the inital metadata for authors

GlynnPhillips:


server/controllers/stream.js
```diff
@@ -110,6 +111,12 @@ module.exports = (req, res, next) => {
 			res.set('Surrogate-Key', res.locals.surrogateKeys.join(' '))
 		}
 	})
+	// Create schema.org microdata
```
GlynnPhillips: Thanks @andygnewman I have moved that now

PR #1561, opened 2016-12-13
Apply search filters

laurieboyes:
@georgecrawford @wheresrhys 

E.g. search 'china' with filters for topic 'Pokemon Go' and org 'Nintendo':
`/search?q=china&filter=topics:Pokemon%20Go&filter=organisations:nintendo`

Temporary filter UI:
screen shot 2016-12-13 at 14 49 56


server/lib/format-sapi-filters-for-query.js
```diff
@@ -0,0 +1,23 @@
...
+//
+// to: 'people:Tony Blair AND organisations:Labour Party UK'
+//
+module.exports = (filters) => {
+
+	if(!filters || !filters.length) {
+		return '';
+	}
+	return ' ' + filters
+			.map(filter => `${filter.taxonomy}:${filter.name}`)
```
georgecrawford: Should this be `${filter.taxonomy}:"${filter.name}"` as per the SAPI example of `people:"David Cameron"`?

laurieboyes: Ah yeah good spot! Doesn't seem to make a difference but worth keeping it consistent

PR #1577, opened 2016-12-15
Rhys/filter search

wheresrhys:
keyword

@georgecrawford @laurieboyes


client/components/search/tag-suggestions.js
```diff
@@ -25,11 +26,18 @@ export class TagSuggestions extends React.Component {
 		return text.replace(RegExp(regExpEscape(this.state.searchTerm), 'gi'), '<mark>$&</mark>');
 	}
 
+	handleSelect (ev) {
+		ev.stopPropagation();
+		let target = ev.target;
+		while (!target.classList.contains('n-typeahead__link')) {
```
laurieboyes: I think safari crushes our classList dreams: https://polyfill.webservices.ft.com/v2/docs/features/#Element_prototype_classList

laurieboyes: Scratch that, there's just very minor test failure on the polyfill which is causing it to say it's unsupported, we're fine

PR #1589, opened 2016-12-19
chore: add storylines to default transforms

TheoLeanse:
@andygnewman what are your thoughts on this? it means that there is a single (mockett-editable) source of truth to determine those concepts we want to display storylines, and we avoid clogging this file up with separate pieces of config for brexit, us-election, organisations that editorial want storylines on... but there's a tradeoff with running the transform (and making a request to our api) for nearly all streams.

([current list of storyline-streams is here](https://github.com/Financial-Times/next-storylines-api/blob/master/server/utils/export-concept-data.js#L6-L27))


andygnewman:
Would leave as per the config for the AB test period. If (when) it proves successful and we discuss with Editorial how they want to use it, we can then modify how we enable it.

TheoLeanse:
@andygnewman after our stupendous AB test success I'm doing the work to allow editorial to choose what streams show storylines. 
I think this means adding the storyline transform to the defaults.

I've also moved the key developments stuff into the api

TheoLeanse:
@andygnewman 

How's this looking? This should prevent a request to the storylines api for every request to stream.

server/transforms/storylines.js
```diff
@@ -1,37 +1,28 @@
...
-				this.storylineData = getStorylineData(req.params.thingId);
-				const queryVariables = {
-					id: req.params.thingId,
-					typeName: 'KeyDevelopments'
-				};
-				this.listedArticles = fetchGraphqlData(listOfTypeQuery, queryVariables);
-			}
+			if (req.perPage.page !== 1 || !flags.streamPageStorylines) return;
+
+			storylines.checkEnabledForStream(req.params.thingId)
```
andygnewman: Given both of these return a promise, would it be better to do the assignment to this.storylineData at the top level promise?
```
this.storylineData = storylines.checkEnabledForStream
```

TheoLeanse: I think I see what you mean and have updated. lmk if it looks any better.

andygnewman:
Yep, looking better. The way it had been written you may have started resolving the inner promise, before the outer promise had made the assignment.

May be good to put it behind a feature flag so it can be turn on / off-able.

TheoLeanse:
so, it's currently feature-flagged inasmuch as it checks
`req.perPage.page === 1 && flags.streamPageStorylines`
before doing anything in the `pre` block

andygnewman:
Might need to get my eyes tested...

server/lib/get-storyline-data.js
```diff
@@ -0,0 +1,36 @@
...
+function fetchData (id) {
+	const STORYLINES_APP = `ft-next-storylines-api-${process.env.REGION === 'US' ? 'us' : 'eu'}`;
+	return fetch(`https://${STORYLINES_APP}.herokuapp.com/concepts/${id}`, {
+		headers: { 'FT-Next-Backend-Key': process.env.FT_NEXT_BACKEND_KEY }
+	})
+		.then(fetchres.json)
+		.catch(error => logger.warn('Failed getting storylines data', error));
+}
+
+module.exports.checkEnabledForStream = checkEnabledForStream;
```
andygnewman: Why exporting these two functions?

andygnewman: Guess you've done this for testing, which I guess is reasonably valid, but you could have approached the tests as scenarios rather than unit tests.

andygnewman:
Just some non-critical questions about testing strategy, otherwise. 👍

server/lib/get-storyline-data.js
TheoLeanse: yep it was for unit tests. there is a test for the more general behaviour of this code which I think is pretty solid, so I could remove the unit tests and the extra exports, but I don't see the harm in leaving them in? unnecessary LOC I guess.

PR #1591, opened 2016-12-19
update to grid system (rollback)

gvonkoss:
@TheoLeanse @andygnewman 
this was throwing off some centering for the storyline component


TheoLeanse:
+1

PR #1603, opened 2016-12-21
Handle: Hub content when no stories in results

andygout:
cc @andygnewman 

Addresses this Sentry error: https://sentry.io/nextftcom/ft-next-stream-page/issues/195588462/

However (in the scenario of the results array having no content), instead of causing the page to error it will now render a Life & Arts page with only `Columnists`, `Most Read` and `Video` components; maybe it would be preferable for the page to fail and for a cached version to be served (given these errors seem to be short-lived)?


andygout:
These errors have not been seen since 17 Jan 2017 so am closing this PR.

PR #1604, opened 2016-12-21
Amend: remove janet yellen from storylines transform

andygnewman:
cc: @gvonkoss 

Was used for development, but not wanted by editorial for live


gvonkoss:
LGTM

PR #1607, opened 2016-12-22
Migrate deep machine to use the new Next API

bjfletcher:
Along with tests so that hopefully they won't get regressed again.


config/graphql-queries/query.js
```diff
@@ -0,0 +1,53 @@
+const fragments = require('@financial-times/n-teaser').fragments;
```
andygnewman: Can we get a better name for this file than just query?

bjfletcher: @andygnewman aha, that shouldn't have been in... removed

server/lib/top-stories.js
```diff
@@ -1,8 +1,9 @@
-import api from 'next-ft-api-client';
 import itemGenreIs from './item-genre-is';
 import fetchGraphqlData from './fetch-graphql-data';
 import specialReportsHubTopStoriesQuery from '../../config/graphql-queries/special-reports-hub-top-stories';
 import topStoriesList from '../../config/graphql-queries/top-stories-list';
+import topStoriesSearch from '../../config/graphql-queries/top-stories-search';
+import topStoriesDeepMachine from '../../config/graphql-queries/deep-machine';
 import backUpTopStories from '../../config/graphql-queries/back-up-top-stories';
 import leadStoryRelatedContent from '../../config/graphql-queries/lead-story-related-content';
 import decorateItem from '../models/next-api-equivalent-decorator';
```
andygnewman: If all data is now coming from next-api this decorator should no longer be needed.

test/server/lib/top-stories.test.js
```diff
@@ -32,6 +34,38 @@ describe('Top Stories from Next API', () => {
...
+
+		beforeEach(() => {
+			fetchGraphqlDataStub.returns(Promise.resolve(topStoriesSearchFixture));
+		});
+
+		it('returns an object with stories array and listId', () => {
+			return subject.fromApi('aStreamId', { aiCuratePage: 'chronoCuration' })
+			.then(result => {
+				expect(result).to.have.all.keys(['stories', 'listId']);
+				expect(result.stories).to.be.an('array');
```
andygnewman: Would it be good to test that it is actually using the right graphQL query here, otherwise any of the methods of getting top stories would meet these conditions.

test/server/lib/top-stories.test.js
```diff
@@ -32,6 +34,38 @@ describe('Top Stories from Next API', () => {
...
+
+		beforeEach(() => {
+			fetchGraphqlDataStub.returns(Promise.resolve(topStoriesDeepMachineFixture));
+		});
+
+		it('returns an object with stories array and listId', () => {
+			return subject.fromApi('aStreamId', { aiCuratePage: 'mlCuration' })
+			.then(result => {
+				expect(result).to.have.all.keys(['stories', 'listId']);
+				expect(result.stories).to.be.an('array');
```
andygnewman: Same comment as above.

test/server/lib/opinion-and-analysis.test.js
```diff
@@ -36,6 +38,39 @@ describe('Opinion And Analysis', () => {
...
+			context('A search for items', () => {
+
+				beforeEach(() => {
+					fetchGraphqlDataStub.returns(Promise.resolve(fixtureSearchNextApi));
+				});
+
+				it('returns an object with stories and a list id', () => {
+					return opinionAndAnalysis.fromApi('XXXXXXXX', { aiCuratePage: 'chronoCuration' })
+						.then(result => {
+							expect(result).to.have.all.keys(['stories', 'listId']);
```
andygnewman: Check that the right graphql query is being executed

test/server/lib/opinion-and-analysis.test.js
```diff
@@ -36,6 +38,39 @@ describe('Opinion And Analysis', () => {
...
+			context('A list using deep machine', () => {
+
+				beforeEach(() => {
+					fetchGraphqlDataStub.returns(Promise.resolve(fixtureDeepMachineNextApi));
+				});
+
+				it('returns an object with stories and a list id', () => {
+					return opinionAndAnalysis.fromApi('XXXXXXXX', { aiCuratePage: 'mlCuration' })
+						.then(result => {
+							expect(result).to.have.all.keys(['stories', 'listId']);
```
andygnewman: Ditto

server/lib/top-stories.js
bjfletcher: They're all now in /dev/null

test/server/lib/top-stories.test.js
bjfletcher: Great, I've updated the unit tests to check which queries are used.

PR #1609, opened 2016-12-30
Remove references to deleted flag

lc512k:
Nothing urgent, but the flag is gone, so cleaning up any references to it


andygnewman:
👍

PR #1634, opened 2017-01-19
Amend: remove all featured podcasts stuff

andygnewman:
cc: @i-like-robots 

This gets rid of the featured  / latest podcasts at the top of the podcast hub page and the show pages.

Also removes the config carve outs for podcasts, result of which is that they match the default transforms and get the full default set rather than just the basic default set (could mean they get trending topics and / or a top section if editorial decided to curate a list for a podcast).


i-like-robots:
CC @mmockett

i-like-robots:
Thank you!

PR #1646, opened 2017-02-01
oFtIcons has been deprecated in favour of oIcons, the builds were fail…

andrewgeorgiou1981:
…ing as the deprecated package was unavailable. Switched to use oIcons
@andygnewman


andygnewman:
👍

PR #1655, opened 2017-02-09
Promo banner on stream page for weekend print / premium digital sub

aintgoin2goa:
@lc512k @keirog 

screen shot 2017-02-09 at 10 35 31


client/components/weekend-print-sub-promo/main.js
```diff
@@ -0,0 +1,26 @@
...
+const CONTENT_CLASS = 'weekend-sub-promo__content';
+const CONTAINER_CLASS = 'weekend-sub-promo';
+
+function onCloseButtonClick (e){
+	e.preventDefault();
+	document.querySelector('.'+CONTAINER_CLASS).remove();
+}
+
+function createCloseLink (){
+	const a = document.createElement('a');
```
keirog: It’d be great if this could be a button instead

client/components/weekend-print-sub-promo/main.js
```diff
@@ -0,0 +1,24 @@
...
+	const button = document.createElement('button');
+	button.classList.add(CLOSE_BTN_CLASS);
+	button.innerHTML = CLOSE_BTN_HTML;
+	button.onclick = onCloseButtonClick;
+	button.setAttribute('data-trackable', 'dismiss');
+	return button;
+}
+
+function init (){
+	document.querySelector('.'+CONTENT_CLASS).appendChild(createCloseButton());
```
keirog: Would be good to be a bit defensive here, as it would break if there wasn’t an element

views/partials/weekend-print-sub-promo.html
```diff
@@ -0,0 +1,17 @@
+<div class="weekend-sub-promo" data-trackable="weekend-print-promo">
```
keirog: Could this div be an `<aside>` instead? Gives a heads up to JAWs users that it’s secondary info

views/partials/weekend-print-sub-promo.html
```diff
@@ -0,0 +1,17 @@
+<aside class="weekend-sub-promo" data-trackable="weekend-print-promo">
+	<div class="o-grid-container">
+		<div class="o-grid-row">
+			<div class="weekend-sub-promo__content">
+				<h2 class="weekend-sub-promo__header">
```
lc512k: In the wider context of the page, this should probably not be a H2 (screen readers would pick it up as _the_ most important section of content under `Life & Arts`)

aintgoin2goa: Even though it's inside an <aside>?  Also, if this is the case you should know I copied the markup from the myft digest component so this problem already exists on all other stream pages...

views/partials/weekend-print-sub-promo.html
```diff
@@ -2,10 +2,10 @@
 	<div class="o-grid-container">
 		<div class="o-grid-row">
 			<div class="weekend-sub-promo__content">
-				<h2 class="weekend-sub-promo__header">
```
lc512k: I meant to say this shouldn't be a H at all (H4 will mean we've skipped from H1 to H4 which is also an infringement). Sorry!

aintgoin2goa: ok 😄

PR #1659, opened 2017-02-14
Rhys/no react

wheresrhys:
@andygnewman


andygnewman:
Not even sure it would need preact in the dependencies?

wheresrhys:
yep needs it for the new search filter

PR #1661, opened 2017-02-15
Fix the display of the weekend print banner

tavvy:
- import css correctly for section pages
- correctly check if showLifArtsSubPromo config is already false


jkerr321:
:+1:

PR #1667, opened 2017-02-21
add advanced search link

bjfletcher:
This adds a link to the advanced search.

image


andygnewman:
👍  Not the best looking link, but does the job for the temporary measure

PR #1669, opened 2017-02-21
NFT-742, bug fix for Life & Arts stream images in IE

andrewgeorgiou1981:
Lazy-loading the hero images does not work in IE and causes layout issues. See NFT-742.
The first image should not be lazy-loaded in any case as it will almost certainly be in view when the user first visits the page.

@bjfletcher


andrewgeorgiou1981:
@bjfletcher @andygnewman

bjfletcher:
You're very right. The fix looks good to me. Yup, let's see if @andygnewman sees something we don't.

PR #1672, opened 2017-02-22
Ads - adding right-rail ad for streams with no tredning-topics

andrewgeorgiou1981:
@bjfletcher @andygnewman 
See ads-1137 - ads were missing from Right-rail.


andygnewman:
Looks like the right logic 👍

PR #1673, opened 2017-02-23
NFT-742: lazy-loading hero image does not work in IE and we should no…

andrewgeorgiou1981:
…t lazy-load the image as it is most likely in view when the user hits the page


andrewgeorgiou1981:
@bjfletcher

bjfletcher:
:tada: Thumbs up

PR #1674, opened 2017-02-23
bring in lists mixins from n-ui

leggsimon:


leggsimon:
Think this can be closed now, right @adgad ?

PR #1675, opened 2017-02-23
add logging to storylines data-fetch

TheoLeanse:
@andygnewman 
adding a log to try to diagnose some weird storylines behaviour...


andygnewman:
👍 OK, happy to chat about the weird issues.

PR #1678, opened 2017-02-23
Make the build green maybe it works now who knows

adgad:
- [x] myft-hint and myft-digest head-css not working?
- [x] o-buttons now no longer imported in n-ui, but is used by myft-digest (in n-ui) - should it just import it?


client/main-hub.scss
```diff
@@ -1,19 +1,9 @@
...
-		myft: true,
-		myftDigestPromo: true,
-		myftHint: true
-	),
-	$preset: 'complete'
-);
-@import 'n-ui/bootstrap';
-
-@include nUiCriticalStart();
+$o-buttons-is-silent: false;
```
wheresrhys: in latest beta of n-ui the myft bits and pieces take care of o-buttons, so possibly worth trying without explicitly including buttons here

adgad: Tried it, hub page still missing styles for the myft digest. All the other pages look okay, so I guess some component on those pages that aren't on the hub page is including o-buttons in non-silent mode.

PR #1680, opened 2017-02-27
Dd/sfb 11

ddigulescu:


bjfletcher:
@ddigulescu This is great. :) Nice work.

In future PRs, it's ok to use a shorter version of the bug description ("Getting '0' items on 2nd page when having more than 1 topic and having 25-50 search results") in the title, instead of bug number.")... for example "search filters: fix broken 2nd search page"

So, go ahead and merge... and give yourself a good pat on the back :)

PR #1684, opened 2017-03-01
Fix ie11 validation

roland-vachter:
Fix for "[IE 11 & IE Mobile 11] Getting validation error when either selecting "last x days/month" or manually inputting dates in the from-to fields."


client/components/search/main.js
```diff
@@ -42,18 +42,59 @@ const validateDateRange = ({from, to}) => {
...
 	if (fromDate < MIN_DATE || toDate > now || fromDate > toDate) {
-		return false
+		return false;
 	}
 
 	return true;
 }
 
+const normalizeDateString = (dateString) => {
+	const dateRegex = /^(\d{2})\/(\d{2})\/(\d{4})$/;
```
bjfletcher: Have you checked that the day and month are padded with 0's if in single digit? I.e., is it 01/02/2017 and not 1/2/2017?
and also that the year will always be 4 digits? not 2 digits? I.e., 2017 and not 17

roland-vachter: The required pattern is DD/MM/YYYY, that's shown to the user as well. Anything else is considered invalid. And I didn't change that, just fixed the validation for it.

PR #1688, opened 2017-03-02
Fix same day search, both visually and the results as well (search on "to" date inclusive!)

roland-vachter:
Fix for "Filtering on a particular day does not work. eg: From:13/02/2017 To:13/02/2017"

The fix was to visually show the date in which the results are filtered, and increased "to" with 24 hours (it means that the search is performed on the "to" date inclusive).


roland-vachter:
@wheresrhys do you mean the search controller file? Actually I didn't have to really dig into it, just searched the references to dateFrom and dateTo, therefore had to change just the 2 simple file, one that formats visually the query and the other one that generates the query itself.

wheresrhys:
ah no, I meant the client side code (it shows how well I understand the codebase if I can't even tell if you're making server or client side changes 😝 )

roland-vachter:
Yeah, well, this change is pure server-side :) I can add some comments from what I understood, but it has nothing to do with this PR :)

roland-vachter:
Apparently this fixes another issue: "Custom date filter set to current date does not appear as a filter tag"

PR #1691, opened 2017-03-03
Filter link hidden in core browsers.

roland-vachter:
Fix for the issue "FIlter link shown on Core browsers (e.g. IE9, Safari iOS 7) - see acceptance criteria SF.20"


views/partials/sub-header/search-filter.html
```diff
@@ -1,4 +1,4 @@
-<section data-trackable="search-filtering" class="search-filtering">
+<section data-trackable="search-filtering" class="search-filtering o--if-js">
```
wheresrhys: would prefer to use this https://github.com/Financial-Times/n-ui-foundations/blob/master/util/main.scss#L37

roland-vachter: Changed.

PR #1694, opened 2017-03-06
Fix for able to add the same Keyword filter (sfb.9)

ddigulescu:
I created a hoc for tag suggestions that receives the container element for already selected tags and returns the original tag suggestions component with the selected tags as props.


client/components/search/tag-suggestions.js
```diff
@@ -65,12 +60,13 @@ export class TagSuggestions extends React.Component {
 	}
 
 	render () {
-		const hasTags = this.state.suggestions.tags && this.state.suggestions.tags.length;
-		const suggestions = this.state.suggestions.tags
+		const hasTags = this.props.suggestions.tags && this.props.suggestions.tags.length;
+		const suggestions = this.props.suggestions.tags
+			.filter(tag => !_.includes(this.props.selectedTags, tag.name))
```
bjfletcher: I'd use `Array.prototype.includes` instead of adding a dependency on `lodash` here.
A good thumb of rule is, if it is listed in here, then we should use it: https://cdn.polyfill.io/v2/docs/features/

ddigulescu: I agree, dropped the lodash dep :)

PR #1697, opened 2017-03-08
Fix styling of the live-stream notice

keirog:
Before | After
---|---
before | after

See individual commits for marginally more context.

The hover/focus state is pretty subtle, but it’s consistent with o-buttons `¯\_(ツ)_/¯`


keirog:
@debugwand

PR #1698, opened 2017-03-09
Fix for triggering on submit listeners when removing a filter

ddigulescu:
User specified date removed when user removes topic filter


client/components/search/main.js
```diff
@@ -362,7 +362,7 @@ module.exports = {
 						formEl.parentNode.removeChild(formEl);
 					}
 				}
-				form.submit();
+				form.querySelector('button[type=submit]').click();
```
roland-vachter: Out of curiosity, what does it change to click the submit button instead of submitting the form?

ddigulescu: It triggers listeners for submit event... .submit() doesn't :). Working with DOM is depressing...

bjfletcher: I suggest adding as a comment above this line:
```
// The submit event only fires when it is activated by a user - and does not fire when activated by code
// Hence this "click" trick. Please see:
// http://wayback.archive.org/web/20090323062817/http://blogs.vertigosoftware.com/snyholm/archive/2006/09/27/3788.aspx
```

PR #1703, opened 2017-03-09
Fix lost stories from special reports streams

debugwand:


debugwand:
the default limit was actually higher so there were 2 data items for the first column that never got shown because of the way the handlebars templates specifies which items to show

PR #1707, opened 2017-03-13
Revert "Update babel presets, plugins and configuration"

commuterjoy:
This reverts commit 2b510d860d3d13c800aae652c0dba8ac00bc7d60.


i-like-robots:
Ta muchly

commuterjoy:
just in case i get woken at 4am :)

i-like-robots:
I was definitely confident we'd be fine because master runs fine on my home laptop.

😨

i-like-robots:
Issues should be fixed in #1708

PR #1715, opened 2017-03-16
split up styles a bit

wheresrhys:


adgad:
It looks okay to me. One thing I'm wondering is what happens if there's duplication in the individual stylesheets?

i.e. hypothetically if `components/teasers` and `components/n-image` both included say o-grid, would they be created separately and then concat'ed inline? Which means it would output two o-grids?

Generally I've been checking /world, /news-feed, fastft, random tag stream, life and arts and /special-reports to QA.

client/main.scss
```diff
@@ -0,0 +1,125 @@
...
+@include nUiStylesheetStart('o-expander');
+	$o-expander-is-silent: false;
+	@import 'o-expander/main';
+@include nUiStylesheetEnd('o-expander');
+
+@include nUiStylesheetStart('search');
+	@import 'components/search/main';
+@include nUiStylesheetEnd('search');
+
+@include nUiStylesheetStart('photo-diary');
```
adgad: Any idea where these actually exist on the site? I can't seem to find them (so possibly a place for further optimisation!)

wheresrhys: I'll have a dig around

wheresrhys:
> what happens if there's duplication in the individual stylesheets?

Good question. It's a risk/consideration when splitting head/main css too. In general duplication of styles is to be avoided given our current tooling for css splitting, and we are being quite good at doing that I think.

This PR seems to not break anything. I'll try a prod build locally to double check

i-like-robots:
Have checked this out and it builds fine. All looks good except for a slightly naughty right rail.

wheresrhys:
closing in favour of https://github.com/Financial-Times/next-stream-page/pull/1721 as circle is being weird

PR #1727, opened 2017-03-21
Delete: remove layouts and sections

andygnewman:
cc: @i-like-robots 

Some housekeeping - removal of legacy of n-card / n-section days.


i-like-robots:
\0/

PR #1728, opened 2017-03-23
Change bg color of search filter tags on focus

ddigulescu:
as suggested by Laura Carvajal


bjfletcher:
Looks good, Dragos... let's go!

